PR #35, opened 2019-11-13
Changing to github actions

Jorropo:
It no longer have `scripts/ci.sh` it use `actions-rs/toolchain` and `actions-rs/cargo` instead.

To enable windows build follow the line `.github/workflows/rust.yml#L40` (2 lines to delete and one to uncoment).


YaronWittenstein:
@Jorropo you are the man!

Issue #21, opened 2019-10-03
Create a basic CLI for SVM


YaronWittenstein:
# Overview / Motivation
We'd like to have a CLI for SVM. It'll help us exercise SVM functionality as a black-box with various configurations and inputs.

# The Task
1. The CLI should support deploying of a new contract.
The input can be provided as one blob of data or field-by-field explicitly.

2. The CLI should support the capability of executing transactions of an existing contract.
The input can be provided as one blob of data or field-by-field explicitly 

# Implementation Notes
1. This file contains documentation of the binary format for deploying a new contract:
https://github.com/spacemeshos/svm/blob/develop/crates/svm-contract/src/wire/deploy/mod.rs

We'd like the CLI to receive the following fields:
* `name` (string)
* `contract code`: file path (string)
* `author address`: a string representing 20 bytes in hexadecimal format (`0xA18C...`)

2. And, this file contains documentation of the binary format for execution a new contract:
https://github.com/spacemeshos/svm/blob/develop/crates/svm-contract/src/wire/exec/mod.rs

We'd like the CLI to receive the following fields:
* `contract address`: a string representing 20 bytes in hexadecimal format (`0xA18C...`)
*  `sender address`: a string representing 20 bytes in hexadecimal format (`0xA18C...`)
* `function`: a string
* `args`: we'd guide you how to represent it as a CLI input

3. Please use the Rust [clap](https://github.com/clap-rs/clap) crate for implementing the command-line:

# Contribution Guidelines
Important: Issue assignment to developers will be by the order of their application and proficiency level according to the tasks complexity. We will not assign tasks to developers who haven't introduced themselves on our Gitter [dev channel](https://gitter.im/spacemesh-os/svm)

1. Introduce yourself on `svm` [dev chat channel](https://gitter.im/spacemesh-os/svm) - ask our team any question you may have about this task
2. Fork branch `develop` to your own repo and work in your repo
3. The code must be formatted using `rust fmt`
4. You must write unit tests for all types and methods when submitting a component.
5. When ready for code review, submit a PR from your repo back to branch `develop`
6. Attach relevant issue to PR

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 1.0 ETH (193.41 USD @ $193.41/ETH)  attached to it as part of the spacemeshos fund.__

[...]

PR #42, opened 2019-11-28
WIP: add cli to deploy and call contract

zoek1:
**Description**
Add create with a cli that allow deploy and call contracts on the svm

**Reference/Fix**
#21


YaronWittenstein:
@zoek1 
we've decided to postpone this task for now...
It should be relevant in the near future.

PR #92, opened 2020-03-01
Support encoding/decoding via CLI tool

moshababo:
Closing https://github.com/spacemeshos/svm/issues/76.


moshababo:
@YaronWittenstein after adding some tests, [one test case](https://github.com/spacemeshos/svm/pull/92/files#diff-1cd123c01f776d7842379ed034ffe49cR27) which involves non-ASCII chars is failing (due to a minor escape char diff). I thought you should have a look at it.

crates/svm-cli/src/app_template.rs
```diff
@@ -0,0 +1,74 @@
...
+    let file = File::open(code_path);
+    let mut file = match file {
+        Ok(v) => v,
+        Err(e) => {
+            let e = format!("failed to open file at {}: {}", code_path, e.to_string());
+            return Err(e.into());
+        }
+    };
+
+    let mut buffer = Vec::new();
```
YaronWittenstein: can you please rename it to `buf`?

crates/svm-cli/src/app_template.rs
```diff
@@ -0,0 +1,74 @@
...
+    let file = File::open(data_path);
+    let mut file = match file {
+        Ok(v) => v,
+        Err(e) => {
+            let e = format!("failed to open file at {}: {}", data_path, e.to_string());
+            return Err(e.into());
+        }
+    };
+
+    let mut buffer = Vec::new();
```
YaronWittenstein: please rename to `buf` here too

crates/svm-cli/tests/app_template_tests.rs
```diff
@@ -0,0 +1,164 @@
+extern crate regex;
```
YaronWittenstein: I believe the usage of `extern crate` is redundant since we use `Rust 2018`

crates/svm-cli/tests/app_template_tests.rs
```diff
@@ -0,0 +1,164 @@
...
+extern crate tempfile;
```
YaronWittenstein: same.

crates/svm-cli/tests/app_template_tests.rs
```diff
@@ -0,0 +1,164 @@
...
+        "myprog",
+        "encode",
+        "app-template",
+        version,
+        name,
+        page_count,
+        code_path,
+        output_path,
+    ];
+    println!("input: {:?}", input);
```
YaronWittenstein: please remove the `println!` from the tests

crates/svm-cli/tests/app_template_tests.rs
```diff
@@ -0,0 +1,164 @@
...
+}
+
+#[test]
+fn decode_invalid_datapath() {
+    let data_path = "non_existing_path";
+    let input = vec!["myprog", "decode", "app-template", data_path];
+    println!("input: {:?}", input);
+
+    let matches = cli::new_app().get_matches_from(input);
+    match cli::process(matches) {
```
YaronWittenstein: you might want to use `if-let`

```rust
if let Err(e) = cli::process(matches) {
   ...
}
else {
  panic!()
}
```

or use an `assert`

```rust
let res = cli::process(matches);
assert!(res.is_err());  // will `panic` when `res` is `Ok(..)`
res.map_err(|e| ...)
```

crates/svm-cli/tests/app_template_tests.rs
```diff
@@ -0,0 +1,164 @@
...
+        Ok(_) => panic!(),
+        Err(e) => assert!(e
+            .to_string()
+            .starts_with(&format!("failed to open file at {}", data_path))),
+    }
+}
+
+fn test_encode_decode(case: AppTemplateTestCase) {
+    let wasm_example_path = &gen_wasm_example_path();
```
YaronWittenstein: I'd personally use a shorter name here

crates/svm-cli/tests/app_template_tests.rs
```diff
@@ -0,0 +1,164 @@
...
+    let input = vec!["myprog", "decode", "app-template", data_path];
+    println!("input: {:?}", input);
+
+    let matches = cli::new_app().get_matches_from(input);
+    let output = cli::process(matches).unwrap();
+    println!("output: {}", output);
+
+    let re = Regex::new(r"Version: (.*)\nName: (.*)\nCode: (.*)\n#Pages: (\d+)").unwrap();
+    let cap = re.captures(&output).unwrap();
+    assert_eq!(&cap[1], case.version);
```
YaronWittenstein: please try to use the just-released `slice patterns` feature,
More info here:
https://blog.rust-lang.org/2020/03/12/Rust-1.42.html

crates/svm-cli/tests/app_template_tests.rs
```diff
@@ -0,0 +1,164 @@
...
+        Err(e) => assert!(e
+            .to_string()
+            .starts_with(&format!("failed to open file at {}", data_path))),
+    }
+}
+
+fn test_encode_decode(case: AppTemplateTestCase) {
+    let wasm_example_path = &gen_wasm_example_path();
+
+    let mut wasm_example_code = Vec::new();
```
YaronWittenstein: I believe you can simplify it by using `include_str!` :
https://doc.rust-lang.org/std/macro.include_str.html

crates/svm-cli/tests/app_template_tests.rs
moshababo: I changed it to a plain const instead, because calling to `fs::canonicalize` wasn't necessary. 
Didn't change the name though.

crates/svm-cli/tests/app_template_tests.rs
moshababo: `include_str!` macro is limited to string literal arg, which I rather not use (because the value is used multiple times).

crates/svm-cli/tests/app_template_tests.rs
moshababo: I didn't find this new feature usable/suitable for this use-case.

PR #98, opened 2020-03-16
Refactor build scripts

moshababo:
* Rust build script to output `svm.h` into `svm/target/release` instead of `svm/target/debug/build/svm-runtime-c-api-X/out`
* `build.sh` to copy objectfile instead of moving
* `build.sh` to copy `svm.h`, created by the Rust build script

### DIFF
--- a/crates/svm-runtime-c-api/build.rs
+++ b/crates/svm-runtime-c-api/build.rs
@@ -1,6 +1,6 @@
 extern crate cbindgen;
 use cbindgen::{Builder, Language};
-use std::{env, fs, path::PathBuf};
+use std::{env, path::PathBuf};
 
 fn main() {
     generate_svm_header();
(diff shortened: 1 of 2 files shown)


crates/svm-runtime-c-api/build.rs
```diff
@@ -1,18 +1,22 @@
...
+
+    // targeting the workspace 'target/release' dir
     let mut src_header = PathBuf::from(&out_dir);
-    src_header.push("svm");
-    src_header.set_extension("h");
+    src_header.pop();
+    src_header.pop();
+    src_header.pop();
+    src_header.pop();
+    src_header.push("release/svm.h");
```
YaronWittenstein: `release` means `production`. please try to use the `PROFILE` cargo env variable.
It seems that if you omit one `src_header.pop()` statement you can have `src_header.push("svm.h")`

moshababo: I thought to always put it under `release`, even if i’m coming from `debug` dir.
If to inspect `PROFILE`, what should be done in case of `debug`? put it under `debug` dir, or not do anything?

YaronWittenstein: I think the best will be to always generate `svm.h` into `PROFILE`.
Then each `example` can decide where to copy its assets from.

PR #105, opened 2020-03-23
cli: encode/decode SpawnApp tx

moshababo:
Closing https://github.com/spacemeshos/svm/issues/77.

WIP: docs, tests (cli, `WasmValue::try_from(&str)`)

### DIFF
--- a/crates/svm-app/src/types/wasm_value.rs
+++ b/crates/svm-app/src/types/wasm_value.rs
@@ -1,3 +1,7 @@
+use crate::types::wasm_value::WasmValueError::InvalidFormat;
+use std::convert::TryFrom;
+use std::str::FromStr;
+
 /// Wasm Integer.
 #[derive(Clone, PartialEq, Debug)]
 pub enum WasmValue {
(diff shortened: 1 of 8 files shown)


crates/svm-app/src/types/wasm_value.rs
```diff
@@ -7,3 +11,47 @@ pub enum WasmValue {
     /// A 64-bit integer.
     I64(u64),
 }
+
```
YaronWittenstein: I'm ambivalent about whether this is the right place for this piece.
I know it's targeted for the SVM CLI but besides that, I don't see any useful scenario for this.

I'd rather move it to the `svm-cli` crate.

crates/svm-cli/src/cli.rs
```diff
@@ -13,21 +12,21 @@ pub fn new_app<'a, 'b>() -> App<'a, 'b> {
                 .about("Encode various transaction types")
                 .setting(AppSettings::SubcommandRequiredElseHelp)
                 .subcommand(
-                    App::new("app-template")
```
YaronWittenstein: I personally love using `_` instead of `_` but it's OK with both.

crates/svm-cli/src/spawn_app.rs
```diff
@@ -0,0 +1,120 @@
...
+            let e = format!(
+                "failed to decode hex string '{}': {}",
+                ctor_buf_hex,
+                e.to_string()
+            );
+            return Err(e.into());
+        }
+    };
+
+    let mut ctor_args_vals: Option<Vec<WasmValue>> = None;
```
YaronWittenstein: please rename it as `ctor_args`.

crates/svm-cli/src/spawn_app.rs
```diff
@@ -0,0 +1,120 @@
...
+    if let Some(args) = ctor_args {
+        let mut vals: Vec<WasmValue> = Vec::new();
```
YaronWittenstein: the `Vec<WasmValue>` seems redundant.

crates/svm-cli/src/spawn_app.rs
```diff
@@ -0,0 +1,120 @@
...
+        let mut errs: Vec<String> = Vec::new();
```
YaronWittenstein: the `Vec<String>` seems redundant.

crates/svm-cli/src/spawn_app.rs
```diff
@@ -0,0 +1,120 @@
...
+
+        ctor_args_vals = Some(vals);
+    }
+
+    let mut builder = SpawnAppBuilder::new()
+        .with_version(version)
+        .with_template(&template_addr.into())
+        .with_ctor_index(ctor_idx)
+        .with_ctor_buf(&ctor_buf);
+    if let Some(vals) = ctor_args_vals {
```
YaronWittenstein: can be written as:
`if let Some(ctor_args) = ctor_args { ... }`

crates/svm-app/src/types/wasm_value.rs
```diff
@@ -7,3 +11,47 @@ pub enum WasmValue {
...
+            return Err(InvalidFormat);
+        }
+        let mid = len - TYPE_CHARS_LEN;
+        let num: String = value.chars().take(mid).collect();
+        let num_t: String = value.chars().skip(mid).collect();
+
+        match &num_t[..] {
+            "i32" => {
+                match u32::from_str(&num) {
+                    Ok(v) => return Ok(WasmValue::I32(v)),
```
YaronWittenstein: the `return` seems redundant.

crates/svm-cli/src/cli.rs
moshababo: I didn't understand what you mean.. you mentioned the same char twice.

crates/svm-cli/src/spawn_app.rs
moshababo: A different name is needed since i'm using `Vec<&str>` to populate `Vec<WasmValue>`.

PR #109, opened 2020-03-28
cli: encode/decode SpawnApp tx

moshababo:
https://github.com/spacemeshos/svm/pull/105 followup.


crates/svm-cli/src/spawn_app.rs
```diff
@@ -0,0 +1,124 @@
...
+            }
+        }
+        if errs.len() > 0 {
+            return Err(format!("invalid ctor_args: {:?}", errs).into());
+        }
+
+        ctor_args_vals = Some(vals);
+    }
+
+    let mut b = SpawnAppBuilder::new()
```
YaronWittenstein: I prefer `builder` over `b`

crates/svm-cli/src/wasm_value.rs
```diff
@@ -0,0 +1,72 @@
...
+        assert_eq!(
+            parse_str("1073741824i32").unwrap(),
+            WasmValue::I32(1073741824)
+        );
+        assert_eq!(
+            parse_str("1073741824i64").unwrap(),
+            WasmValue::I64(1073741824)
+        );
+
+        assert_eq!(parse_str("").err().unwrap(), ParseError::InvalidFormat);
```
YaronWittenstein: You can extract the reoccurring pattern:
```
parse_str(XXX).err().unwrap()
```

into a helper function (even inside `test_parse_str` test):

```rust
#[test] 
fn test_parse_str() {
   fn to_wasm_value(s: &str) -> WasmValue {
      // ...
   }
}
```

crates/svm-cli/src/spawn_app.rs
```diff
@@ -0,0 +1,124 @@
...
+
+use crate::wasm_value;
+use svm_app::{
+    raw::{decode_spawn_app, NibbleIter},
+    testing::SpawnAppBuilder,
+    types::{SpawnApp, WasmValue},
+};
+use svm_common::Address;
+
+pub fn encode(
```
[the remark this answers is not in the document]
moshababo: Done, but I called them `decode_XXX` since they parse strings into native data structures (used for the later encoding).

crates/svm-cli/src/wasm_value.rs
moshababo: I did it slightly different - please check.

crates/svm-cli/src/wasm_value.rs
```diff
@@ -0,0 +1,74 @@
...
+
+#[derive(Debug, PartialEq)]
+pub enum ParseError {
+    InvalidFormat,
+    UnsupportedType(String),
+}
+
+/// Potentially converts `&str` to `WasmValue`.
+/// Expected format is `{num}{type}` (`type` in (i32|i64)).
+pub fn parse_str(value: &str) -> Result<WasmValue, ParseError> {
```
YaronWittenstein, line 13: a more idiomatic way would be to implement:

```rust
use std::convert::TryFrom;

impl TryFrom<&str> for WasmValue {
  type Error = ParseError;

  fn try_from(value: &str) -> Result<Self, Self::Error> {
     # ...
  }
}
```

crates/svm-cli/tests/spawn_app_tests.rs
```diff
@@ -0,0 +1,216 @@
...
+        version,
+        template_addr,
+        ctor_index,
+        ctor_buf,
+    ];
+
+    let matches = cli::new_app().get_matches_from(input);
+    let res = cli::process(matches);
+    assert!(res.is_err());
+    assert!(res
```
YaronWittenstein, line 211: please try to assert against a `struct/enum` that represents the error:
something like:

```rust
assert_eq!(Err(DecodeError::InvalidDigit)), res);
```

crates/svm-cli/tests/spawn_app_tests.rs
```diff
@@ -0,0 +1,216 @@
...
+        output_path,
+        version,
+        template_addr,
+        ctor_index,
+    ];
+
+    let matches = cli::new_app().get_matches_from(input);
+    let res = cli::process(matches);
+    assert!(res.is_err());
+    assert!(res
```
YaronWittenstein, line 123: please try to assert against a struct/enum that represents the error:
something like:

```rust
assert_eq!(Err(DecodeError::InvalidDigit)), res);
```

crates/svm-cli/tests/spawn_app_tests.rs
```diff
@@ -0,0 +1,216 @@
...
+        "myprog",
+        "encode",
+        "spawn_app",
+        output_path,
+        &case.version,
+        &case.template_addr,
+        &case.ctor_index,
+        &case.ctor_buf,
+    ];
+    for i in 0..case.ctor_args.len() {
```
YaronWittenstein, line 53: It seems the loop could be replaced with:

```rust
input.extend_from_slice(&case.ctor_args[..]);
```

crates/svm-cli/tests/spawn_app_tests.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    let re = Regex::new(
+        r"Version: (.*)\nTemplate: (.*)\nctor_idx: (\d+)\nctor_buf: (.*)\nctor_args: (.*)",
+    )
+    .unwrap();
+    let caps = re.captures(&output).unwrap();
+
+    assert_eq!(&caps[1], case.version);
+    assert_eq!(&caps[2], {
+        let template_addr = hex::decode(case.template_addr).unwrap();
```
YaronWittenstein, line 76: better to introduce some helper function (maybe a local function within the test).

```rust
let template_addr = normalize(..);
assert_eq!(&caps[2], template_addr);
```

crates/svm-cli/tests/spawn_app_tests.rs
```diff
@@ -0,0 +1,216 @@
...
+        let template_addr = Address::from(template_addr.as_slice());
+        template_addr.fmt(4, 4, " ")
+    });
+    assert_eq!(&caps[3], case.ctor_index);
+    assert_eq!(
```
YaronWittenstein, line 81: same. please extract the normalization out of the `assert_eq!`

crates/svm-cli/tests/spawn_app_tests.rs
```diff
@@ -0,0 +1,216 @@
...
+            hex::decode(case.ctor_buf)
+                .unwrap()
+                .iter()
+                .take(4)
+                .collect::<Vec<_>>()
+        )
+    );
+    assert_eq!(
+        &caps[5],
+        format!("{:?}", {
```
YaronWittenstein, line 94: same.

crates/svm-cli/tests/spawn_app_tests.rs
```diff
@@ -0,0 +1,216 @@
...
+    assert_eq!(&caps[1], case.version);
+    assert_eq!(&caps[2], {
+        let template_addr = hex::decode(case.template_addr).unwrap();
+        let template_addr = Address::from(template_addr.as_slice());
+        template_addr.fmt(4, 4, " ")
+    });
+    assert_eq!(&caps[3], case.ctor_index);
+    assert_eq!(
+        &caps[4],
+        format!(
```
YaronWittenstein, line 83: same.

crates/svm-cli/tests/app_template_tests.rs
```diff
@@ -82,18 +125,19 @@ fn encode_invalid_outputpath() {
...
-            .to_string()
-            .starts_with(&format!("failed to create file at {}", output_path))),
-    }
+    let res = cli::process(matches);
+    assert!(res.is_err());
+    assert!(res
+        .err()
+        .unwrap()
+        .to_string()
+        .starts_with(&format!("failed to create file at {}", output_path)));
```
YaronWittenstein, line 134: please try to assert against a struct/enum that represents the error:
something like:

```rust
assert_eq!(Err(CreateFileError { .. } )), res);
```
The best error here is probably `std::io::Error`:
https://doc.rust-lang.org/std/io/struct.Error.html

crates/svm-cli/tests/app_template_tests.rs
```diff
@@ -55,12 +97,13 @@ fn encode_invalid_codepath() {
...
-            .to_string()
-            .starts_with(&format!("failed to open file at {}", code_path))),
-    }
+    let res = cli::process(matches);
+    assert!(res.is_err());
+    assert!(res
+        .err()
+        .unwrap()
+        .to_string()
+        .starts_with(&format!("failed to open file at {}", code_path)));
```
YaronWittenstein, line 106: please try to assert against a struct/enum that represents the error:
something like: 

```rust
assert_eq!(Err(FileOpenError { ..} ), res);
```

The best error here is probably `std::io::Error` (for `FileNotFound`)
https://doc.rust-lang.org/std/io/struct.Error.html

crates/svm-cli/src/wasm_value.rs
moshababo: This was my go-to, but I got err `E0117`:
```
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types

   | impl TryFrom<&str> for WasmValue {
   | ^^^^^-------------^^^^^---------
   | |    |                 |
   | |    |                 `svm_app::types::wasm_value::WasmValue` is not defined in the current crate
   | |    `str` is not defined in the current crate
   | impl doesn't use only types from inside the current crate
   |
   = note: define and implement a trait or new type instead
```
I tried a few workarounds including using type alias and type redefine, without any success.

crates/svm-cli/tests/spawn_app_tests.rs
moshababo: The returned err type is currently `String`.
Do you mean I should create a custom struct/enum err type, or return an err of the `hex::FromHexError` type?
I didn't use either of them because the err returned from `cli::process` need to support other types as well (e.g., `io::Error`), so I thought to use `String` as the baseline format.

crates/svm-cli/tests/spawn_app_tests.rs
moshababo: see https://github.com/spacemeshos/svm/pull/109#discussion_r404111736

crates/svm-cli/tests/spawn_app_tests.rs
moshababo: Due to the `String`/`&str` conversion, the only kind of solution I found for this is:
```rust
let ctor_args: Vec<_> = case.ctor_args.iter().map(|s| s.as_str()).collect();
input.extend_from_slice(&ctor_args);
```
Which I think is more complicated IMO.
Let me know what you think.

crates/svm-cli/tests/app_template_tests.rs
moshababo: see https://github.com/spacemeshos/svm/pull/109#discussion_r404111736

crates/svm-cli/tests/app_template_tests.rs
moshababo: see https://github.com/spacemeshos/svm/pull/109#discussion_r404111736

crates/svm-cli/src/wasm_value.rs
YaronWittenstein: my bad, `WasmValue` is defined in another crate.

crates/svm-cli/tests/spawn_app_tests.rs
moshababo: Actually, I might be able to return different err types by using a trait type, but AFAIK i'm unable to prepend additional string information to the err.

YaronWittenstein: It's not that critical, but since you are the owner of the API, you can have an ad-hoc
error type (`struct / enum`) to make tests more explicit and less brittle.

crates/svm-cli/tests/spawn_app_tests.rs
YaronWittenstein: I've overlooked the `ctor_arg` contains `&str` (and not `String`).
AFAIK you suggestion is more idiomatic than writing a loop.

crates/svm-cli/tests/spawn_app_tests.rs
moshababo: I'm just wandering what's the best practice for this situation.

YaronWittenstein: I personally prefer having an explicit `Error` type over generic `Box<dyn Error>`

moshababo: Ok, i'll fix this in a follow-up PR, because the scope of this is lager than `SpawnApp`.

PR #122, opened 2020-04-16
Fix multisig wallet WASM build

moshababo:
The compiler must produce a dynamic system library (`cdylib`).

### DIFF
--- a/apps/src/wallet/build.sh
+++ b/apps/src/wallet/build.sh
@@ -1,7 +1,7 @@
 cargo +nightly build --target wasm32-unknown-unknown --release
 cp target/wasm32-unknown-unknown/release/*.wasm ../../../apps/wasm
 
-# compress the wasm artifact
+# optimize the wasm artifact in order to reduce size
 pushd ../../wasm
 wasm2wat wallet.wasm > wallet.wast
 rm wallet.wasm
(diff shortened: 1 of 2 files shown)


apps/src/wallet/build.sh
```diff
@@ -1,5 +1,7 @@
 cargo +nightly build --target wasm32-unknown-unknown --release
 cp target/wasm32-unknown-unknown/release/*.wasm ../../../apps/wasm
+
+# compress the wasm artifact
 pushd ../../wasm
 wasm2wat wallet.wasm > wallet.wast
```
YaronWittenstein, line 6: @moshababo you need to run also
`wat2wasm` back again to get the reduced size `wasm`

apps/src/wallet/build.sh
```diff
@@ -1,5 +1,7 @@
 cargo +nightly build --target wasm32-unknown-unknown --release
 cp target/wasm32-unknown-unknown/release/*.wasm ../../../apps/wasm
+
+# compress the wasm artifact
```
YaronWittenstein, line 4: I'd not use the word `compress`.
`optimize` or `removing unused code` would better describe the process I think.

apps/src/wallet/build.sh
moshababo: This is already happening, I haven't change that. 
You might need to expand github's code viewer.

PR #127, opened 2020-05-03
Failing test on Ubuntu

YaronWittenstein:
# Motivation

Trying to figure out why there is a failing test only on Ubuntu when Rust is compiled in `release` mode.


YaronWittenstein:
Added issue: #141

PR #138, opened 2020-05-24
Runtime creates wasm memory

YaronWittenstein:
# Motivation

This PR removes the `Buffer` primitive from `SVM` and uses plain `Wasm Memory` instead.

Future PRs:
* Limit how much memory should be allocated per wasm instance.
  See also:
  https://github.com/wasmerio/wasmer/issues/1360


crates/svm-runtime/src/runtime/default.rs
```diff
@@ -355,21 +355,16 @@ where
...
-    fn init_instance_buffer(&self, func_buf: &Vec<u8>, instance: &mut wasmer_runtime::Instance) {
-        const ARGS_BUF_ID: u32 = 0;
-
+    fn copy_func_buf_to_memory(&self, func_buf: &[u8], instance: &mut wasmer_runtime::Instance) {
         let ctx = instance.context_mut();
+        let memory = ctx.memory(0);
 
-        helpers::buffer_create(ctx.data, ARGS_BUF_ID, func_buf.len() as u32);
+        let func_size = func_buf.len();
+        let view = &memory.view::<u8>()[0..func_size];
```
moshababo, line 363: What about bound checking for the memory size?

crates/svm-runtime/src/runtime/default.rs
```diff
@@ -458,7 +453,7 @@ where
         func_index
     }
 
-    fn prepare_args_and_memory(&self, tx: &AppTransaction) -> Vec<wasmer_runtime::Value> {
+    fn prepare_func_args(&self, tx: &AppTransaction) -> Vec<wasmer_runtime::Value> {
         debug!("runtime `prepare_args_and_memory`");
```
moshababo, line 457: `prepare_args_and_memory` -> `prepare_func_args` ?

crates/svm-runtime/tests/runtime_tests.rs
```diff
@@ -325,3 +325,61 @@ fn runtime_exec_app_reaches_oog() {
...
+    let receipt = runtime.spawn_app(&bytes, &creator, HostCtx::new(), maybe_gas);
+    assert!(receipt.success);
+
+    let app_addr = receipt.get_app_addr();
+    let init_state = receipt.get_init_state();
+
+    // 3) executing an app-transaction
+    let func_idx = 1;
+    let func_buf = vec![0x10, 0x20, 0x30, 0x40, 0x50, 0x60, 0x70, 0x80];
+    let func_buf_ptr = WasmValue::I32(0);
```
moshababo, line 369: Can you give more explanations how this works (in conjunction with the `runtime_func_buf.wast` code)?

crates/svm-runtime/src/runtime/default.rs
YaronWittenstein: this should not happen here.
when we allocate memory we know the size of the `func_buf`

crates/svm-runtime/tests/runtime_tests.rs
YaronWittenstein: I have an idea of how to make it easier to understand.

I prefer to open a new PR that will add two new vmcalls (that we'll need)
and change the test to be easier

crates/svm-runtime/src/runtime/default.rs
moshababo: But you're currently don't check it there - do you mean that you will add that check later, or that you'll verify that `func_buf` size is less than 1 page size?

crates/svm-runtime/src/runtime/default.rs
```diff
@@ -359,6 +359,17 @@ where
         let ctx = instance.context_mut();
         let memory = ctx.memory(0);
 
+        // we don't need to do bounds-checking since we know the instance memory
+        // is larger than the `func_buf`.
+        //
+        // Each wasm instance memory contains at least one `WASM Page`. (A `Page` size is 64KB)
+        // The `len(func_buf)`
```
moshababo, line 366: This line is not clear.

crates/svm-runtime/src/runtime/default.rs
YaronWittenstein: * The `validate_template` will check that `func_buf` isn't big enough.
* The `alloc_wasmer_memory` will have to allocate enough memory for the program to run.

I've added an issue: #140

PR #151, opened 2020-06-07
External KV Store for App's Storage 

YaronWittenstein:
# Motivation

Since `SVM` underlying kv-store for managing App's Storage will be implemented by `go-spacemesh` - we should come up with some FFI interface between `SVM` and `go-spacemesh`.

Upon `SVM Runtime` initialization, FFI functions pointers will be given to `SVM`. 
Then, `SVM` will wrap them using a native Rust interface.

Note: `SVM` will have a separate key-value store for managing the `Template`(s) and `App`(s).
This kv-store will have nothing to do with the FFI one.

SMIP Issue:
https://github.com/spacemeshos/SMIPS/issues/20


crates/svm-storage/src/kv/ffi.rs
```diff
@@ -1,80 +1,129 @@
...
+pub type DiscardFn = unsafe extern "C" fn();
+
+/// # Checkpoint
+///
+/// Notifies the `Host` to checkpoint key-value `State`.
+/// Returns the `State` derived for the checkpoint.
+///
+/// Computing a checkpoint doesn't guarantee that the pendind changes and checkpoint
+/// have been persisted. It's up to the `Host` to determine when to save data for long-term usage.
+pub type CheckpointFn = unsafe extern "C" fn(*mut u8);
```
moshababo, line 42: Is this the same as `commit`? if so, why using this term? (checkpoint/discard instead commit/rollback)

I thought that checkpoint is about rewind-related optimization.

crates/svm-storage/src/kv/ffi.rs
```diff
@@ -1,80 +1,129 @@
...
+///
+/// Notifies the `Host` to discard the current executed transaction changes.
+pub type DiscardFn = unsafe extern "C" fn();
+
+/// # Checkpoint
+///
+/// Notifies the `Host` to checkpoint key-value `State`.
+/// Returns the `State` derived for the checkpoint.
+///
+/// Computing a checkpoint doesn't guarantee that the pendind changes and checkpoint
```
moshababo, line 40: typo: pendind

crates/svm-storage/src/kv/ffi.rs
YaronWittenstein: I've decided not to use the term `commit` here since I wanted to stress that it's up to the `Host` to decide when
to do the `committing/flushing` to disk.

So `checkpoint` here means: 
```
"I'm SVM has completed my part in executing the transaction.
I'd signaling the Host that I'm done and I'd like to know my new State.
It's up to you the Host, to determine when you want to persist the changes."
```

PR #152, opened 2020-06-08
Modifications to key-value abstractions

YaronWittenstein:
# Motivation

Since it has been decided that SVM's State will be managed by `go-spacemesh`, the current key-value abstractions should undergo some changes.

* There will be a distinction between a raw key-value store to stateful key-value store.

* Flushing pending changes will occur only after all layer's transactions have been executed.
The new abstraction should take that into consideration.


crates/svm-kv/src/traits.rs
```diff
@@ -1,9 +1,16 @@
-/// `KVStore` is a trait for defining an interface against key-value stores, for example `rocksdb/leveldb`.
-pub trait KVStore {
-    /// Retrieves the value pointed by `key` (Optional).
+/// A trait for defining an interface against a raw key-value store (e.g `leveldb/rocksdb`).
+pub trait RawKV {
+    /// # Gets the `value` pointed by by `key`.
+    ///
+    /// If there's a matchikg pending `value` (i.e not flushed yet) it should be returned.
```
moshababo, line 5: typo: matchikg

crates/svm-storage/src/kv/traits.rs
```diff
@@ -1,12 +1,78 @@
...
+    /// when calling the `flush` method.
+    ///
+    /// Subsequent `get` calls on `key` should return the new set `value`.
+    ///
+    /// See also: `flush`
+    fn set(&mut self, key: &[u8], value: &[u8]);
+
+    /// Discards all pending changes
+    ///
+    /// This functionality is useful when an execetud application transaction has failed.
```
moshababo, line 30: typo: execetud

PR #153, opened 2020-06-14
new crates: `svm-types` and `svm-codec`

YaronWittenstein:
# Motivation

We're breaking the `svm-app` crate that logically contains:

* Many common types and transaction types.
* Encoding & decoding of SVM transactions

This PR will introduce two new crates:

* `svm-types`: will contain the most types declared under `svm-app`.
* `svm-codec`: will contain the encoding & decoding of transactions.
This crate should also compile to a WASM file and be used by the Spacemesh Wallet
(as part of synthesizing binary transactions to be submitted to the network).

### Future PRs:

* Depercating `svm-app` crate (using `svm-codec` instead).
Other code of `svm-app` will move to the `svm-runtime` crate.

* A likely future code change after this PR will be moving the code of the `svm-common` crate to sit under `svm-types`


crates/svm-codec/examples/test.js
```diff
@@ -0,0 +1,232 @@
...
+}
+
+function instanceCall(instance, func_name, buf) {
+    const func = instance.exports[func_name];
+    return func(buf)
+}
+
+function wasmNewBuffer(instance, object) {
+    const objectStr = JSON.stringify(object);
+    const bytes = new TextEncoder('utf-8').encode(objectStr);
```
moshababo, line 23: All tests are failing for me locally with `ReferenceError: TextEncoder is not defined`. 

Which Nodejs version do you use? it should be specified in `package.json`. 
Mine:
```
Moshes-MacBook-Pro:examples moshababo$ node -v
v10.15.3
```

If this is problematic, maybe you should use the well-supported `Buffer` class instead (assuming this code is intended to run in Nodejs and not in a browser).

crates/svm-codec/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+/// ## WASM Deploy-Template
+///
+/// Reads the WASM buffer given at parameter `buf_ptr` containing a JSON value.
+/// Encodes a `deploy-template` binary-transaction using that JSON value.
+///
+/// Returns a pointer to a new WASM buffer holding the encoded transaction.
+/// If the encoding failed, the returned WASM buffer will contain a JSON with the error.
+#[no_mangle]
+#[cfg(target_arch = "wasm32")]
+pub extern "C" fn wasm_deploy_template(buf_ptr: i32) -> i32 {
```
moshababo, line 50: Also here I found it confusing - we're not deploying a template here, but encoding an app template. 
So maybe `wasm_encode_app_template` instead?

Same for the other types.

crates/svm-codec/examples/run.sh
```diff
@@ -0,0 +1,4 @@
+./../build.sh 
+cp ../../../target/wasm32-unknown-unknown/debug/svm_codec.wasm svm_codec.wasm
+
+npm test
```
moshababo, line 4: This test need to be added to the CI.

crates/svm-codec/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+use varuint14::{decode_varuint14, encode_varuint14};
+use version::{decode_version, encode_version};
+
+/// ## WASM Deploy-Template
+///
+/// Reads the WASM buffer given at parameter `buf_ptr` containing a JSON value.
+/// Encodes a `deploy-template` binary-transaction using that JSON value.
+///
+/// Returns a pointer to a new WASM buffer holding the encoded transaction.
+/// If the encoding failed, the returned WASM buffer will contain a JSON with the error.
```
moshababo, line 47: From reading this, it's not clear how the OK/ERROR result type can be detected. So perhaps you should add a brief explanation about that.

crates/svm-codec/examples/package.json
```diff
@@ -0,0 +1,28 @@
...
+  ],
+  "author": "Yaron Wittenstein",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/spacemeshos/svm/issues"
+  },
+  "homepage": "https://github.com/spacemeshos/svm#readme",
+  "dependencies": {
+    "mocha": "^8.0.1",
+    "typescript": "^3.9.5"
```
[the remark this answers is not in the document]
YaronWittenstein: I'll remove it. I initially thought to code in typescript

crates/svm-codec/examples/test.js
YaronWittenstein:
```
❯ node -v
v12.6.0
```

YaronWittenstein: I'll add to `package.json`

crates/svm-codec/src/template/wire.rs
```diff
@@ -0,0 +1,120 @@
...
+
+use crate::{
+    decode_varuint14, decode_version, encode_varuint14, helpers, Field, NibbleIter, NibbleWriter,
+    ParseError,
+};
+
+use svm_layout::{DataLayout, DataLayoutBuilder};
+
+/// Encodes a raw Deploy-Template.
+pub fn encode_deploy_template(template: &AppTemplate, w: &mut NibbleWriter) {
```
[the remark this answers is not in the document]
YaronWittenstein: https://github.com/spacemeshos/svm/issues/158
@moshababo

crates/svm-codec/src/lib.rs
YaronWittenstein: https://github.com/spacemeshos/svm/issues/158

PR #161, opened 2020-07-20
Receipts to json

YaronWittenstein:
# Motivation

The motivation of this PR is to decode `Receipt`(s) given as raw data into a friendly JSON.
This feature is required by the `Explorer` and `smapp` projects.

While working on the PR, I've also made changes to the way Receipts errors are encoded.
Now, there is a single enum `ReceiptError` that gathers all the Receipts errors (formerly we had an error struct per each Receipt type)

@avive @lrettig  FYI


crates/svm-codec/src/api/json/receipt.rs
```diff
@@ -0,0 +1,353 @@
...
+    let mut map: serde_json::Map<String, Value> = std::mem::take(map);
+
+    map.insert("type".into(), Value::String(ty.into()));
+    map.insert("success".into(), Value::Bool(false));
+    map.insert("logs".into(), Value::Array(logs));
+
+    map.into()
+}
+
+fn decode_deploy_template(receipt: &TemplateReceipt, ty: &'static str) -> Value {
```
moshababo, line 115: If you're receiving the native data type and returning a JSON, then you are encoding, not decoding, no?

crates/svm-codec/src/api/json/receipt.rs
```diff
@@ -0,0 +1,353 @@
+use serde_json::{json, Value};
+
+use crate::api::json::{self, JsonError};
+use crate::api::raw;
+
+use svm_types::receipt::{
+    ExecReceipt, Log, ReceiptError, ReceiptOwned, SpawnAppReceipt, TemplateReceipt,
+};
+
+pub fn decode_receipt(json: &Value) -> Result<Value, JsonError> {
```
moshababo, line 10: Same here; you're receiving a raw version, decode it to native (`raw::decode_receipt`), and then encode it to JSON. If so, the function should be called `encode_receipt`.

crates/svm-codec/src/api/json/receipt.rs
YaronWittenstein: I'm expecting an input JSON looking like this:

```json
{
  "data":  'SEE EXPLANATION BELOW' 
}
```

The `data` entry expects in the value **encoded Receipt given as a string**  

For example, The 0xA12D (2 bytes) should be provided as 'A12D' (4 String chars).
So if the original encoded Receipt consumes `N` bytes, the `data` value should be exactly `2 * N` chars.

PR #168, opened 2020-08-16
calldata

YaronWittenstein:
# Motivation

This PR makes a fundamental change to the way data is passed to a Smart-Contract 
(`spawn-app` or `exec-app` SVM transactions).

Instead of having both `func_args` and `func_buf` this PR will deprecate them in favor of a new mechanism called `calldata`. The `calldata` is an extension of the old `func_buf` but it will introduce new primitives. 

Also, the new `calldata` won't use the `svm-nibble` crate at all. 
It means the encoded data will be slightly less packed but it will make it easier to re-implement the encoding/decoding of `calldata` in other languages in the future. 

A future PR will implement the `returndata`. It will enable Receipts of executed transactions
to return arbitrary data. This data will have to be decoded by the client.


crates/svm-abi/encoder/src/types/array.rs
```diff
@@ -0,0 +1,40 @@
...
+
+        let marker = match self.len() {
+            0 => layout::ARR_0,
+            1 => layout::ARR_1,
+            2 => layout::ARR_2,
+            3 => layout::ARR_3,
+            4 => layout::ARR_4,
+            5 => layout::ARR_5,
+            6 => layout::ARR_6,
+            7..256 => layout::ARR_0_255,
```
moshababo, line 23: Does this mean that each array with size > 6 will be considered as size = 255?

crates/svm-abi/layout/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+//! | 0 | 1 1 1 | 0 1 0 1 |  i64 (unsigned) - 8 bytes |
+//! +---+-------+---------+---------------------------+
+//! | 0 | 0 0 0 | 0 1 1 0 |  Array - 0 items          |
+//! | 0 | 0 0 1 | 0 1 1 0 |  Array - 1 item           |
+//! | 0 | 0 1 0 | 0 1 1 0 |  Array - 2 items          |
+//! | 0 | 0 1 1 | 0 1 1 0 |  Array - 3 items          |
+//! | 0 | 1 0 0 | 0 1 1 0 |  Array - 4 items          |
+//! | 0 | 1 0 1 | 0 1 1 0 |  Array - 5 items          |
+//! | 0 | 1 1 0 | 0 1 1 0 |  Array - 6 items          |
+//! | 0 | 1 1 1 | 0 1 1 0 |  Array - 0..255 items     |
```
moshababo, line 66: Are you keeping the leading bit for something else, or was it just that there was no need to use it? You always separate him from the next 3 bits, so I was wandering.

crates/svm-abi/decoder/src/decoder.rs
```diff
@@ -70,55 +83,171 @@ impl Decoder {
...
     }
 
-    fn verify_array(&self, _value: &Value) -> Result<(), DecodeError> {
-        // todo!()
-        Ok(())
+    fn decode_i64(&self, cursor: &mut Cursor) -> Result<i64, DecodeError> {
+        let byte = self.read_byte(cursor)?;
+
+        let nbytes = match byte {
+            layout::I64_1B | layout::U64_1B => 1,
```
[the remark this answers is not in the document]
YaronWittenstein: the encoding will consume less space

crates/svm-abi/encoder/src/types/array.rs
YaronWittenstein: no, it means, that I'll add a `length` field to the encoding

crates/svm-abi/layout/src/lib.rs
YaronWittenstein: yes, the MSB bit stands for `has more`

YaronWittenstein: that way we could always had more types.
However, I don't expect that we'll need more than a single byte as it's now

crates/svm-abi/decoder/src/decoder.rs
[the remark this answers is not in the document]
moshababo: What's the advantage over using smaller size types? And if there is one, why also support smaller size types?

crates/svm-abi/decoder/src/decoder.rs
```diff
@@ -70,55 +83,171 @@ impl Decoder {
...
+            TypeKind::Amount => self.decode_amount(cursor)?.into(),
+            TypeKind::I8 => self.decode_i8(cursor)?.into(),
+            TypeKind::U8 => self.decode_u8(cursor)?.into(),
+            TypeKind::I16 => self.decode_i16(cursor)?.into(),
+            TypeKind::U16 => self.decode_u16(cursor)?.into(),
+            TypeKind::I32 => self.decode_i32(cursor)?.into(),
+            TypeKind::U32 => self.decode_u32(cursor)?.into(),
+            TypeKind::I64 => self.decode_i64(cursor)?.into(),
+            TypeKind::U64 => self.decode_u64(cursor)?.into(),
+            TypeKind::Array => self.decode_array(cursor)?,
```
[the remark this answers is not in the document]
YaronWittenstein: we'll add a Blob type when we'll need

crates/svm-abi/encoder/src/types/array.rs
moshababo: It's still not clear to me, can you elaborate more?

crates/svm-abi/decoder/src/decoder.rs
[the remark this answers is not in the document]
YaronWittenstein: if you have `i64` with a smaller value you can encode it in less than 8 bytes.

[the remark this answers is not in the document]
YaronWittenstein: You still encode `i64` type. The underlying encoding is different (it's transparent to the user)

crates/svm-abi/encoder/src/types/array.rs
YaronWittenstein: you can encode `i64` in many ways according to the value.

if the value fits a single-byte you prepend it with a `i64-one-byte` marker
if the value fits a two-byte you prepend it with a `i64-twt-byte` marker 
and so on....

moshababo: So if the value fits a single-byte, why not just use `i8` instead? What's the reasons for preferring `i64`?

moshababo: Also, it still doesn't answer the question regarding the Array size.

YaronWittenstein: > So if the value fits a single-byte, why not just use `i8` instead? What's the reasons for preferring `i64`?

if you have a function that expects an `i64` you will call it multiple times with a different value.
when you call it with a small value the encoding will be different from when you'll call it with bigger values.
In both cases, we'll encode an `i64` type. The exact encoding method will be different

crates/svm-abi/decoder/src/decoder.rs
```diff
@@ -70,55 +83,171 @@ impl Decoder {
...
+        let byte = self.read_byte(cursor)?;
+        let nitems = match byte {
+            layout::ARR_0 => 0,
+            layout::ARR_1 => 1,
+            layout::ARR_2 => 2,
+            layout::ARR_3 => 3,
+            layout::ARR_4 => 4,
+            layout::ARR_5 => 5,
+            layout::ARR_6 => 6,
+            layout::ARR_0_255 => self.read_byte(cursor)?,
```
YaronWittenstein, line 238: @moshababo here we read a byte to know what's the length of the array.
We know it's between 6 to 255 (inclusive) - but this byte will tell us the exact length.

crates/svm-abi/encoder/src/types/array.rs
YaronWittenstein: > Also, it still doesn't answer the question regarding the Array size.

Similarly to the `i64` example.
You encode Arrays. Different inputs can result in completely different implementations.
For Array's larger than 6 - we have `length` field (a single byte)
For Array's larger than 256 (someday, if relevant) - we might want to have another encoding.

PR #183, opened 2020-10-21
`endpoint` proc-macro

YaronWittenstein:
# Motivation

Adding `[endpoint]` proc-macro attribute. It should ease the implementation of apps' endpoints.

This macro implements:

* Auto-parsing of the input calldata to Rust native types (see PR: #181)

* Will generate in a future PR the API schema in a JSON format (will be consumed by `smapp`). For now, it will gather the required data in compile-time.


crates/svm-sdk/macros/tests/endpoint.rs
```diff
@@ -0,0 +1,119 @@
...
+        fn test<F>(f: F) where F: FnOnce() {
+            // run the tests in a serial manner
+            let guard = TEST_LOCK.lock().unwrap();
+
+            let host = MockHost::instance();
+            host.reset();
+
+            f();
+
+            drop(guard)
```
moshababo, line 46: Isn't `guard` going to be dropped here anyway?

crates/svm-sdk/macros/tests/storage.rs
```diff
@@ -0,0 +1,271 @@
+// use cfg_if::cfg_if;
+
```
moshababo, line 2: Why commented?

crates/svm-sdk/storage/src/mock.rs
```diff
@@ -173,7 +173,7 @@ fn storage() -> MutexGuard<'static, InnerStorage> {
 mod tests {
     use super::*;
 
-    use crate::memory::alloc;
+    use svm_sdk_alloc::alloc;
```
moshababo, line 176: `STORAGE` instance init/access pattern is different than `HOST` - If there's no reason for that, it should be unified.

crates/svm-sdk/alloc/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+//! Using this crate when writing SVM Templates in Rust isn't mandatory but should be very useful.
+//!
+//! The crate is compiled with `![no_std]` (no Rust stdlib) annotation in order to reduce the compiled WASM size.
+
+#![allow(missing_docs)]
+#![allow(unused)]
+#![allow(dead_code)]
+#![allow(unreachable_code)]
+
+extern crate alloc;
```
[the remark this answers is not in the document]
YaronWittenstein: I prefer to keep the current name for now.
I don't expect it to change in the future.

moshababo (approved):

Issue #129, opened 2020-05-17
add `capacity` to `svm_byte_array`


YaronWittenstein:
# Motivation

In case the `bytes` pointer of `svm_byte_array` points to data of a `Vec`,
we need in the Rust side to use [`Vec::from_raw_parts`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts)

This is required mainly for managing `Receipts`.

Since the `length` of a `Vec` might be strictly less than its `capacity`, it's better not to assume anything about the relation of these two fields and add explicit `capacity` field to the `svm_byte_array`.

YaronWittenstein:
Using the new [`Vec#leak`][Vec#leak] instead of [`Vec#into_raw_parts`][Vec#from_raw_parts] makes this issue not necessary. 

[Vec#leak]: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.leak
[Vec#from_raw_parts]: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts

PR #185, opened 2020-11-07
add `capacity` to `svm_byte_array`

moshababo:
Added `capacity` to `svm_byte_array` (the total number of allocated bytes).
It may be unequal and bigger than `length` if the `svm_byte_array` instance is an alias to an instance of a data structure such as `Vec` (which in order to properly get deallocated needs first to be re-constructed using the proper allocated capacity).

[Vec#leak](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.leak) wasn't used because in order to remove the surplus capacity it clones the data, while `svm_byte_array` is intended to be an alias. 

Closing https://github.com/spacemeshos/svm/issues/129.

### DIFF
--- a/crates/svm-runtime-c-api/src/byte_array.rs
+++ b/crates/svm-runtime-c-api/src/byte_array.rs
@@ -16,7 +16,7 @@ use std::{convert::TryFrom, string::FromUtf8Error};
 /// ```
 ///
 #[allow(non_camel_case_types)]
-#[derive(Copy, Clone)]
+#[derive(Clone)]
 #[repr(C)]
 pub struct svm_byte_array {
     /// Raw pointer to the beginning of array.
@@ -139,7 +139,7 @@ mod tests {
         assert_eq!(s1_len, bytes.length);
         assert_eq!(s1_capacity, bytes.capacity);
 
-        let s2 = String::try_from(bytes).unwrap();
+        let s2 = String::try_from(bytes.clone()).unwrap();
         assert_eq!(s2, "Hello World!".to_string());
         assert_ne!(s2.as_ptr(), bytes.bytes); // `s2` is a clone.
         assert_eq!(s2.len() as u32, bytes.length);
--- a/crates/svm-runtime-c-api/src/byte_array.rs
+++ b/crates/svm-runtime-c-api/src/byte_array.rs
@@ -9,12 +9,10 @@ use std::{convert::TryFrom, string::FromUtf8Error};
 /// use svm_runtime_c_api::svm_byte_array;
 ///
 /// let s1 = "Hello World!".to_string();
-/// let ptr = s1.as_ptr();
-/// let length = s1.len() as u32;
-/// let bytes = svm_byte_array { bytes: ptr, length };
+/// let bytes: svm_byte_array = s1.as_bytes().into();
 ///
-/// let s2 = String::try_from(bytes);
-/// assert_eq!(s1, s2.unwrap());
+/// let s2 = String::try_from(bytes).unwrap();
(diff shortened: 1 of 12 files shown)


crates/svm-runtime-c-api/src/byte_array.rs
```diff
@@ -56,31 +61,43 @@ impl Default for svm_byte_array {
...
+/// assert_eq!(s1.len() as u32, bytes.length);
+/// assert_eq!(s1.len() as u32, bytes.capacity);
 ///
-/// let s = String::try_from(array);
-/// assert_eq!("Hello World!".to_string(), s.unwrap());
+/// let s2 = String::try_from(bytes).unwrap();
+/// assert_eq!(s1.to_string(), s2);
 /// ```
 ///
 impl From<&str> for svm_byte_array {
```
YaronWittenstein, line 74: I believe we'll have now to manage without `From<&str>`
since a slice in Rust don't expose a `capacity` (it makes sense since it's an implementation detail)
the new code should have only `From<...>` from types we can extract it's `capacity`.

I'd suggest that you remove it and let's see where the compiler yells.

crates/svm-runtime-c-api/src/byte_array.rs
```diff
@@ -108,20 +125,25 @@ impl TryFrom<svm_byte_array> for String {
 ///
 impl From<&[u8]> for svm_byte_array {
```
YaronWittenstein, line 126: same as `From<&str>` - pleas remove this and let's see the implications.

crates/svm-runtime-c-api/src/macros.rs
```diff
@@ -72,6 +72,7 @@ macro_rules! impl_into_svm_byte_array {
                 $crate::svm_byte_array {
                     bytes: bytes.as_ptr(),
                     length,
+                    capacity: length,
```
YaronWittenstein, line 75: at line `64` above we can extract the real capacity:

```rust
let capacity = bytes.capacity() as u32;
```

crates/svm-runtime-c-api/src/value.rs
```diff
@@ -157,6 +158,7 @@ mod tests {
         let bytes = svm_byte_array {
             bytes: raw.as_ptr(),
             length: raw.len() as u32,
+            capacity: raw.len() as u32,
```
YaronWittenstein, line 161: should be:

```rust
capacity: raw.capacity() as u32,
```

crates/svm-runtime-c-api/src/value.rs
```diff
@@ -170,6 +172,7 @@ mod tests {
...
```
YaronWittenstein, line 175:
```rust
capacity: raw.capacity() as u32,
```

crates/svm-runtime-c-api/src/macros.rs
moshababo: `bytes` is an array (hence without `capacity()`), no?

crates/svm-runtime-c-api/src/byte_array.rs
moshababo: Yes, I was thinking that it's unsafe to continue supporting it, but wasn't sure if we want to keep it anyway.

Compilation is fine after the removal, but some tests are failing. Will fix it.

crates/svm-runtime-c-api/src/macros.rs
YaronWittenstein: I don't remember, if it's then indeed `capacity = length`

YaronWittenstein (approved):

PR #187, opened 2020-11-17
Wasmer 1.0.0 alpha5

YaronWittenstein:
# Motivation

* Updating wasmer to version `1.0.0-alpha5`
* Changing the way external imports (i.e FFI imports) work according to the new wasmer c-api


crates/svm-runtime-c-api/tests/api_tests.rs
```diff
@@ -1,23 +1,179 @@
...
+/// in the import function signature. In addition to required memory for the `results` will be allocated prior calling `trampoline`
+/// such that the `trampoline` will only be left with placing the `results` values.
+///
+/// In case the `trampoline` failed, a pointer to heap-allocated trap will be propagated back to SVM.
+/// SVM will be responsible of deallocating that memory pointed by that `wasm_trap_t`.
+#[no_mangle]
+unsafe extern "C" fn trampoline(
+    env: *mut c_void,
+    args: *const wasm_val_vec_t,
+    results: *mut wasm_val_vec_t,
```
moshababo, line 113: The usage of `wasm_val_vec_t` (in `wasm_func_callback_with_env_t` signature) creates a dependency between `svm` client and `wasmer`'s C API. 
This type is being constructed _after_ `wasmer` is calling `svm`'s inner callback, hence `svm` could use a different type and a different signature for `trampoline`.

This type is defined in `wasmer` using a macro. I'm not sure if importing it will require merely using the C headers, or the lib as well. In any case, this would be suboptimal. 

It seems the most straightforward to use `svm`'s `Vec<WasmValue>` (which is also more restrictive in the supported Wasm types). Its encoding format is already defined and used. We can use it, or define `Vec<T>` and `WasmValue` as C types, similarly to how it's done in `wasmer` C API.

crates/svm-runtime/src/import.rs
```diff
@@ -1,51 +1,169 @@
...
+                    let zero = wasm_val_t::try_from(Val::I64(0)).unwrap();
+                    let mut results: wasm_val_vec_t = vec![zero; num_rets].into();
+
+                    let trap = callback(*env, &processed_args, &mut results);
+
+                    if !trap.is_null() {
+                        let trap: Box<wasm_trap_t> = Box::from_raw(trap);
+
+                        /// TODO: we want access to `trap.inner`
+                        /// (this field has visibility of `pub(crate)`)
```
moshababo, line 100: So we don't have a solution for this yet?

crates/svm-runtime-c-api/tests/api_tests.rs
```diff
@@ -1,23 +1,179 @@
...
+) -> *mut wasm_trap_t {
```
moshababo, line 114: `wasm_trap_t` also has the same problem of creating a dependency with `wasmer`'s C API.

crates/svm-runtime/src/import.rs
YaronWittenstein: no, we need to ask the wasmer team to change that

crates/svm-runtime/src/import.rs
```diff
@@ -1,51 +1,169 @@
...
+    trap::wasm_trap_t,
+    types::{wasm_functype_t, wasm_valkind_enum},
+    value::{wasm_val_t, wasm_val_vec_t},
+};
+
+#[allow(non_camel_case_types)]
+#[derive(Clone)]
+#[repr(C)]
+pub struct svm_env_t {
+    pub inner_env: *const c_void,
```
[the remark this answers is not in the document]
YaronWittenstein: It's a matter of taste. I don't really mind changing that.
The motivation was to keep the API with the `env` naming.

crates/svm-runtime-c-api/tests/api_tests.rs
YaronWittenstein: good point, I've overlooked that fact (I guess I was on an auto-pilot since I'm used to the way things were in the past in the old version of wasmer)

crates/svm-runtime-c-api/tests/api_tests.rs
YaronWittenstein: yes - same

crates/svm-ffi/src/env.rs
```diff
@@ -0,0 +1,57 @@
...
+
+#[allow(non_camel_case_types)]
+#[derive(Clone)]
+#[repr(C)]
+pub struct svm_env_t {
+    pub inner_env: *const c_void,
+
+    pub host_env: *const c_void,
+
+    pub returns: svm_wasm_types_t,
```
moshababo, line 15: What's the purpose of this field? Will it ever need to be used by the extern client?
I see that `svm` use it in `inner_callback` to allocate the `results` buffer, but I don't think it's necessary - why not just read it via closure, instead of doing set/get via `env`?

crates/svm-ffi/src/value.rs
```diff
@@ -31,25 +31,69 @@ use crate::svm_byte_array;
...
+
+    assert!(nvalues <= std::u8::MAX as usize);
+
+    let capacity = 1 + nvalues * 9;
+
+    let mut bytes = Vec::with_capacity(capacity);
+
+    bytes.write_u8(nvalues as u8).unwrap();
+
+    for ty in types {
```
moshababo, line 65: I'd rather have a simpler version of this function, and have it to merely allocate the needed buffer size.
But you can leave as-is if you prefer.

crates/svm-ffi/src/value.rs
```diff
@@ -31,25 +31,69 @@ use crate::svm_byte_array;
...
+/// let values = Vec::<WasmValue>::try_from(&values).unwrap();
+/// assert_eq!(values, vec![WasmValue::I64(0), WasmValue::I32(0), WasmValue::I64(0)]);
+/// ```
+///
+pub fn alloc_wasm_values(types: &[WasmType]) -> svm_byte_array {
+    let nvalues = types.len();
+
+    assert!(nvalues <= std::u8::MAX as usize);
+
+    let capacity = 1 + nvalues * 9;
```
moshababo, line 59: You can potentially reduce the size if you'll scan through the actual types.

crates/svm-ffi/src/byte_array.rs
```diff
@@ -0,0 +1,284 @@
...
+/// let bytes: svm_byte_array = s1.into();
+///
+/// let s2 = String::try_from(bytes).unwrap();
+/// assert_eq!(s2, "Hello World!".to_string());
+/// ```
+///
+#[allow(non_camel_case_types)]
+#[derive(Clone)]
+#[repr(C)]
+pub struct svm_byte_array {
```
moshababo, line 27: Need to be exposed via the C API.

crates/svm-runtime-c-api/tests/api_tests.rs
```diff
@@ -1,23 +1,156 @@
...
+
+const COUNTER_MUL_FN_INDEX: u32 = 123;
+
+fn wasm_trap(err: String) -> *mut svm_trap_t {
+    let trap: svm_trap_t = err.into();
+
+    // this heap-allocated memory will be released by SVM.
+    // (See: `ExternImport#wasmer_export`)
+
+    Box::into_raw(Box::new(trap))
```
moshababo, line 51: Same as previously with `results`, this is expected to be allocated by the extern client, and to be released by `svm`, hence can lead to a crash or a leak.

crates/svm-ffi/src/value.rs
YaronWittenstein: That's right, but I believe it doesn't worth it.
Actually, we could have said from the first place that each value consumes exactly 8 bytes.

crates/svm-ffi/src/byte_array.rs
YaronWittenstein: yes - I'm on it

YaronWittenstein: The only workaround I've found for now is:
1. use `cbindgen` for the `svm-ffi` crate - it should output a C header named `svm_types.h`
This header file will contain the definition for `svm_byte_array` along with other used FFI-structs within `svm-runtime-c-api`

2. The `cbindgen` of `svm-runtime-c-api` will still output the `svm.h` header but it will have an include statement of `include "svm_types.h"` inside.

3. Each successful GitHub Actions CI run will have `svm_types.h` in its generated artifacts.

crates/svm-ffi/src/value.rs
YaronWittenstein: I've added a comment explaining why I prefer that we allocate the worst case.
Since memory allocation is a costly operation we should strive to minimize the number of its usages.

The savings of a few bytes is negligible in this case.

crates/svm-runtime-c-api/tests/api_tests.rs
YaronWittenstein: That's a bit tricky since `SVM` can't know ahead-of-time how much space to allocate for `svm_trap_t`.
The only workaround I've in mind is to agree on a maximum trap size - say 1024 bytes for example.

crates/svm-runtime-c-api/tests/api_tests.rs
```diff
@@ -1,23 +1,179 @@
...
+    args: *const wasm_val_vec_t,
+    results: *mut wasm_val_vec_t,
+) -> *mut wasm_trap_t {
+    let env: &svm_env_t = env.into();
+    let func_idx = env.host_env::<func_index_t>();
+    let callback = func_index_to_callback(func_idx);
+
+    if let Some(callback) = callback {
+        let args = prepare_args(args);
+        let ctx = env.inner_mut();
```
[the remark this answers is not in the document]
YaronWittenstein: The current code base still uses `inner_env` under `svm_env_t`
If you feel it's critical let's chat about that

crates/svm-runtime/src/import.rs
```diff
@@ -1,52 +1,173 @@
...
+pub struct ExternImport {
+    name: String,
 
-    pub returns: Vec<WasmType>,
+    namespace: String,
 
-    pub func_ptr: *const c_void,
+    params: Vec<WasmType>,
+
+    returns: Rc<Vec<WasmType>>,
```
YaronWittenstein, line 20: @moshababo 
I've thought about how to simplify the code of the `results` allocations and still gain performance.

So I've picked using `Rc` to share the `returns` types with the `inner_callback` using shallow clone.
(the runtime overhead is minimal and should be much less expansive than cloning the `Vec`.

crates/svm-runtime/src/import.rs
```diff
@@ -1,52 +1,173 @@
...
+
+                        let err_msg: String = (*trap).into();
+                        let err = RuntimeError::new(err_msg);
+
+                        return Err(err);
+                    }
+
+                    match Vec::<WasmValue>::try_from(&results) {
+                        Ok(vals) => {
+                            // TODO: validate the returns types.
```
YaronWittenstein, line 76: In general, this should be done via the `host trampoline` as part of injecting the `results`.

I'm not sure we want to have this validation logic duplicated here

crates/svm-ffi/src/value.rs
moshababo: That was my comment in https://github.com/spacemeshos/svm/pull/187#discussion_r528360072. In this comment I'm not referring to that, but to the filling of the allocated bytes (with `nvalues`, and `ty` for each value). Although the extern client must comply with the signature, I think it would be better for this function to just allocate the buffer.

crates/svm-runtime-c-api/tests/api_tests.rs
moshababo: In `go-wasmer` the solution is:
1. Allocate the error in Go
2. Create C-compatible alias that Rust can use (via `to_wasm_trap_new`, defined in cgo preamble)
3. Call Rust's C-API and re-allocate it on the heap  (via `wasm_trap_new`)
4. Go allocation to be automatically released
5. Rust allocation to be released by Go (via `wasm_trap_delete`). I'm not sure about this step, because Rust is using the trap raw point (`trampoline` return value), so it's not clear how would it still be valid (if released by Go immediately after `trampoline` returns), or why Go even need to bother with timing its release. So i'm either missing something, or the code is still in the works.

crates/svm-ffi/build.rs
```diff
@@ -0,0 +1,31 @@
...
+    // build using cbindgen
+    Builder::new()
+        .with_language(Language::C)
+        .with_crate(crate_dir.clone())
+        .with_include_guard("SVM_TYPES_H")
+        .include_item("svm_byte_array")
+        .include_item("svm_trap_t")
+        .include_item("svm_func_callback_t")
+        .include_item("svm_env_t")
+        .with_documentation(false)
```
moshababo, line 27: Why without documentation (unlike `svm-runtime-c-api`)?

crates/svm-ffi/src/byte_array.rs
moshababo: I tested the artifacts - works fine.

crates/svm-runtime/src/import.rs
moshababo: Looks good.

crates/svm-runtime/src/import.rs
moshababo: Yes, can probably avoid doing it here.

crates/svm-runtime-c-api/tests/api_tests.rs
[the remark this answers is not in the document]
moshababo: It's fine, whatever you think.

crates/svm-ffi/src/value.rs
YaronWittenstein: done

crates/svm-runtime-c-api/tests/api_tests.rs
YaronWittenstein: `svm-runtime-c-api` added a method of trap allocation. (also a method for deallocation - in general, it should be used only for tests since SVM is in charge of deallocation of traps)

crates/svm-runtime/src/import.rs
YaronWittenstein: I've added validation of the `results` since SVM only allocates a zeroed-buffer before calling the host function.

crates/svm-runtime/src/import.rs
```diff
@@ -1,52 +1,201 @@
...
-            WasmType::I64 => WasmerType::I64,
+            WasmType::I32 => Type::I32,
+            WasmType::I64 => Type::I64,
             _ => panic!("Only i32 and i64 are supported."),
         })
         .collect()
 }
+
+#[inline]
+fn to_wasm_values(bytes: &svm_byte_array, types: &[WasmType]) -> Option<Vec<WasmValue>> {
```
moshababo, line 154: Wouldn't it be better to return `Result` instead of `Option`, since you later transform the `None` option to `Err` anyway? That way you could return more specific errors from `to_wasm_values`.

crates/svm-runtime-c-api/tests/api_tests.rs
```diff
@@ -1,23 +1,151 @@
...
+/// We use here integers as function identifiers to be called by the `trampoline`.
+#[derive(Debug)]
+#[repr(C)]
+struct func_index_t(u32);
+
+type Callback = fn(&mut Context, &[WasmValue]) -> Result<Vec<WasmValue>, &'static str>;
+
+const COUNTER_MUL_FN_INDEX: u32 = 123;
+
+fn wasm_trap(err: String) -> *mut svm_trap_t {
```
moshababo, line 45: Any reason for still keeping this function?

crates/svm-runtime-c-api/tests/api_tests.rs
```diff
@@ -1,23 +1,151 @@
...
+                /// We copy the values returned by `callback` to `results`.
+                /// This copying operation must not fail (otherwise it's an undefined-behavior).
+                let results: &mut svm_byte_array = &mut *results;
+                results.copy_wasm_values(&values);
+
+                /// since `callback` didn't error, we return a `NULL` pointer signaling
+                // that there no trap has occurred.
+                return std::ptr::null_mut();
+            }
+            Err(err) => wasm_trap(err.to_string()),
```
moshababo, line 107: Instead of `wasm_trap` you should use the new `svm_trap_alloc` C-API function.

crates/svm-runtime-c-api/src/api.rs
```diff
@@ -835,7 +874,24 @@ pub unsafe extern "C" fn svm_runtime_destroy(runtime: *mut c_void) {
 #[must_use]
 #[no_mangle]
 pub unsafe extern "C" fn svm_imports_destroy(imports: *const c_void) {
-    let _ = Box::from_raw(imports as *mut Vec<Import>);
+    let _ = Box::from_raw(imports as *mut Vec<ExternImport>);
+}
+
+/// Allocates a new `svm_trap_t` with inner `error` of size `size`.
+#[no_mangle]
+pub unsafe extern "C" fn svm_trap_alloc(size: u32) -> *mut svm_trap_t {
```
moshababo, line 882: I think that a simpler design would be to accept `svm_byte_array` msg and to re-allocate it on the heap using Rust (similarly to wasmer's `wasm_trap_new`). Copying bytes to an unsafe buffer should work in Go, but it's quite hacky.

crates/svm-runtime-c-api/tests/api_tests.rs
```diff
@@ -1,23 +1,148 @@
...
+    }
+}
+
+unsafe fn prepare_args(args: *const svm_byte_array) -> Result<Vec<WasmValue>, &'static str> {
+    let args: &svm_byte_array = &*args;
+
+    Vec::<WasmValue>::try_from(args).map_err(|_| "Invalid args")
+}
+
+fn wasm_error(msg: String) -> *mut svm_byte_array {
```
moshababo, line 60: Any reason for still using this instead of `svm_wasm_error_create`?

YaronWittenstein: `svm_wasm_error_create` expects `svm_byte_array`.

I can make it work but I prefer to keep the current code as it is.

moshababo (approved):

PR #198, opened 2020-12-20
svm-sdk macros: auto-generate the `svm_alloc` export method.

YaronWittenstein:
# Motivation

Each SVM app's code (a template if to be precise) must export a method named `svm_alloc` of signature: `fn(i32) -> i32`. 

It expects as input the `#bytes` to allocate and it returns a Wasm Memory Pointer to the allocated instance's memory.
This is required since SVM then copies the transaction `calldata` section into that memory region.
(i.e to the `#bytes` memory cells starting from the pointer returned by `svm_alloc`).

Here is where `svm-runtime` is invoking the `svm_alloc`:
https://github.com/spacemeshos/svm/blob/7b6d1ed4d00e9f9e64682849c54db011097c256a/crates/svm-runtime/src/runtime/default.rs#L429

So in order to make sure that `svm-sdk` macros generate SVM compatible Wasm apps we must have it 
inject the `svm_alloc` function (it's a reserved name that can be used as an `#[endpoint], #[ctor] and such).


crates/svm-sdk/src/lib.rs
```diff
@@ -336,15 +336,14 @@ pub use log::log;
...
-pub use svm_sdk_macros::app;
-
 pub use svm_abi_decoder::{CallData, DecodeError, ReturnData};
 pub use svm_sdk_alloc::{alloc, Ptr};
+pub use svm_sdk_macros::app;
+
+// in order to use the following `global allocator` one should
+// call `extern crate svm_sdk;` (instead of `use svm_sdk;`)
+#[global_allocator]
+pub static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
```
moshababo, line 346: It's not clear to me how this should be used. The target exported func should be `svm_alloc`, no? If so, is `ALLOC` being utilized for Rust's allocs?

YaronWittenstein: the usage of `#[global_allocator]` implies that `ALLOC` will be the allocator used.
So it means that `svm_alloc` will use it when it's executing.

moshababo (approved):

PR #205, opened 2021-01-07
Shortened Export name

YaronWittenstein:
# Motivation

Since reducing the volume size of SVM transactions is crucial, the compiled Apps into Wasm will have shortened export names for functions (the `ctor`s and `endpoints). 

For better UX the friendly name will appear in the generated schema (it's off-chain data).

Testing the feature properly will be done in future PR as part of testing a Wasm artifact compiled by the Rust SDK.


moshababo:
The schema tests need to be fixed.

moshababo (approved):

PR #223, opened 2021-02-21
WIP: Adding the `verify` stage

YaronWittenstein:
## Motivation

As part of the Accounts Abstraction effort, this PR adds the so-called `verify` stage into SVM.
Each app will have a built-in `svm_validate` method that will be called by the Miner in order to know whether
the transaction is valid or not. If it's not, then it won't be added to the block (transaction to execute).

Seems PR doesn't assume anything regarding whether or not an app is allowed to have access to its storage
when executing the `verify` method. (the same for any other functionalities that rely on its `State`).

In case, the app has no explicit `svm_validate` we treat the app as if it had one returning always `false` (meaning always `invalid`).

Edit: since the Accounts Abstraction has been put on hold, we didn't finalize yet the spec for the `verify` stage.
(This PR continues the work of #224 and lays the foundations for the `verify`).


tal-m:
Small note: if an app doesn't have a `verify` method, treat it as always `false` rather than always `true`.

YaronWittenstein:
> Small note: if an app doesn't have a `verify` method, treat it as always `false` rather than always `true`.

np

moshababo (approved):

PR #232, opened 2021-04-26
Fixed-Gas 

YaronWittenstein:
# Motivation

Prior to executing a transaction, we should be able to tell what its price will be in terms of gas units.

In the future, we could extend the gas pricing to support metering. Since we only deal with fixed-gas Wasm programs for now - we compute the gas costs offline before actually executing the transaction.

Even though the internals of SVM infers a Gas minimum and maximum range to be paid, in practice the maximum gas units will be charged off the `sender`. 

Using pricing should be integrated into the `Runtime` as part of the validation of a transaction. This will be done in a future PR.

In case a transaction has not enough gas (i.e the `gas limit` is too small) the transaction should be discarded.
(it won't be part of a block and won't get a chance to execute at all).


YaronWittenstein:
Reopened here:
https://github.com/spacemeshos/svm/pull/237

PR #235, opened 2021-05-04
Fixed-Gas Call-Graph

YaronWittenstein:
# Motivation

This PR implements a `Call-Graph` which is derived from running static-analysis over input Wasm programs.
In order for a Wasm program to be considered a Fixed-Gas valid, it must not have any loops or call-cycles.

So if function `f0` has an explicit call to function `f1` which might call in some circumstances to `f0` we have a cycle under our corresponding `Call-Graph.` Of course, a calls-cycle may involve more than two functions as in the example here.

A recursive call (function `f0` calling itself) is treated outside of the `Call-Graph` (it's done while scanning the Wasm code and building the `Call-Graph`).

When the `Call-Graph` has no cycles, it's considered a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph).
The implications of it are that we can use [Topological sorting](https://en.wikipedia.org/wiki/Topological_sorting) in order to determine the processing order of functions for pricing.
(it will be done in a future PR).

When the `Call-Graph` has cycles, the Topological sort will fail. 

This PR implements the Topological sort - and if it reaches a dead-end it continues to move forward and returns a cycle.
 (Note: there might be more than a single cycle under the `Call-Graph`).

Returning a cycle is optional since its purpose is only for debugging or testing.
In production mode, it should be disabled since it's a waste of CPU cycles. It will suffice to report that the input has cycles without giving any specific occurrences.


crates/gas/src/call_graph/sort.rs
```diff
@@ -0,0 +1,140 @@
...
+    // We order so that the tests will be deterministic.
+    // So we always pick the first node out of the sorted `candidates`
+    candidates.sort_by_key(|node| node.value());
+
+    let first = candidates.first().unwrap();
+
+    let mut cycle = Vec::new();
+    let mut visited = HashSet::new();
+
+    let found = find_cycle(&first, &mut cycle, &mut visited);
```
moshababo, line 105: Perhaps re-adjust `find_cycle` signature by having a separate one for the recursion, and one for the external usage.

crates/gas/src/call_graph/builder.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    pub fn add_call(&mut self, source: T, dest: T) {
+        debug_assert!(source != dest);
+
+        self.add_outgoing(source, dest);
+        self.add_incoming(source, dest);
+    }
+
+    pub fn build(self) -> CallGraph<T> {
```
moshababo, line 38: Since building is done on-the-fly, prior to this call, perhaps a different name is more appropriate.

crates/gas/src/call_graph/node.rs
```diff
@@ -0,0 +1,246 @@
...
+use std::hash::{Hash, Hasher};
+use std::rc::Rc;
+
+use super::Value;
+
+/// This struct implement a Node for the `CallGraph`.
+///
+/// Each node has:
+///
+/// * A alue associated with it. This value is assumed to be unique across.
```
moshababo, line 13: You seem to have a typo here, and the connection between the 2 sentences isn't clear enough. Also, you described only the value property, while it looks as if you intended to described all properties.

crates/gas/src/call_graph/node.rs
```diff
@@ -0,0 +1,246 @@
...
+        let value = self.value();
+        let other = other.value();
+
+        value.eq(&other)
+    }
+}
+
+impl<T> Eq for Node<T> where T: Value {}
+
+/// The `NodeRef wrapper struct it adds readability to the code
```
moshababo, line 118: The backtick isn't closed, and also "it" seems redundant.

crates/gas/src/call_graph/node.rs
```diff
@@ -0,0 +1,246 @@
...
+/// (less cognitive load instead of seeing `Rc<RefCell<Node<T>>>` all over the place
+#[repr(transparent)]
+pub struct NodeRef<T> {
+    inner: Rc<RefCell<Node<T>>>,
```
moshababo, line 122: Can you elaborate a bit on the need for using `Rc<RefCell<Node<T>>>`?

crates/gas/src/call_graph/builder.rs
YaronWittenstein: it's a convention in Rust to use `Builder` - i.e we add changes and once we're done we call `build`

crates/gas/src/call_graph/node.rs
YaronWittenstein: It's a common practice in Rust:
https://doc.rust-lang.org/book/ch15-05-interior-mutability.html#having-multiple-owners-of-mutable-data-by-combining-rct-and-refcellt

crates/gas/src/call_graph/node.rs
```diff
@@ -0,0 +1,246 @@
...
+        let other = other.value();
+
+        value.eq(&other)
+    }
+}
+
+impl<T> Eq for Node<T> where T: Value {}
+
+/// The `NodeRef wrapper struct it adds readability to the code
+/// (less cognitive load instead of seeing `Rc<RefCell<Node<T>>>` all over the place
```
moshababo, line 119: The parentheses isn't closed.

crates/gas/src/call_graph/node.rs
```diff
@@ -0,0 +1,250 @@
...
+        let other = other.value();
+
+        value.eq(&other)
+    }
+}
+
+impl<T> Eq for Node<T> where T: Value {}
+
+/// The `NodeRef` wrapper struct adds readability to the code
+/// (less cognitive load instead of seeing `Rc<RefCell<Node<T>>>` all over the place
```
moshababo, line 119: The parentheses aren't closed.

PR #256, opened 2021-07-05
Add a CI lint via cargo fmt

### LINKED ISSUE #255: Automating formatting tasks and lints for Rust files
As described in `CONTRIBUTING.md`, developers are expected to only ever commit well-formatted code (i.e. running `rustfmt` after every change). However, at the time of writing, `master` features 5 unformatted files and we'd like to catch these style issues automatically in the future.

# Possible solutions

The way I see it, we can either run lints on CI using GitHub workflows or writing a Git pre-commit hook. Several points to take into consideration when dealing with pre-commit hooks:

- They mess with the developer's personal development environment, which is big no-no and best avoided.
- They add a `rustfmt` dependency for *all* contributors, even for simple typo corrections.
- Formatting *still* should be verified server-side.

As such, I think we should simply add a lint check on the CI pipeline and call it a day.

neysofu:
Resolves #255. No Git pre-commit hook is planned.


neysofu:
CI times seem not to be impacted as installing `rustfmt` takes little time.

YaronWittenstein:
cool!

PR #258, opened 2021-07-07
Unroll ergonomics

neysofu:
This is a PR draft to discuss *loop unrolling ergonomics* in the context of the SVM ABI encoder. Future changes and/or PRs might address the SVM decoder in a similar fashion if we're satisfied with these changes. The underlying tone of the refactor is:

- Replace the overwhelming number of macro-based `impl`'s with type safety using Rust traits. We might have to tweak trait names and implementation algebra a bit more, but I think this looks better and is much simpler to deal with.
- Unify trait implementations for numeric primitives using the `num-traits` crate.
- Unroll loops using the `seq-macro` crate.
- Prune lots of code branches from the encoder logic by finding recurring patterns.

Possible TODOs and further changes to discuss:

- [x] Naming of layout utility generators: which crate they go in and how to integrate them within the existing codebase.
- [ ] More comments and internal documentation.
- [x] Evaluate similar decisions for the SVM ABI decoder


crates/abi/encoder/src/types/numeric.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use svm_abi_layout::layout;
+
+    #[test]
+    fn integer_layouts() {
```
YaronWittenstein, line 115: If you can - please add a few more asserts to the other integer markers

crates/abi/encoder/src/types/numeric.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use crate::traits::{Numeric, Push};
+use crate::{ByteSize, Encoder};
+
+impl<T, W> Encoder<W> for T
+where
+    T: Numeric + ByteSize + num_traits::PrimInt,
+    W: Push<Item = u8>,
+{
+    fn encode(&self, w: &mut W) {
```
YaronWittenstein, line 12: great piece!

crates/abi/encoder/src/types/numeric.rs
neysofu: I'm not sure I understand, what other conditions should I test?

YaronWittenstein: There are more markers that can be asserted

neysofu: They are derived automatically form the `1B` ones using arithmetic. I think all are covered

Issue #260, opened 2021-07-07
Remove `unsafe` where it makes sense, add SAFETY comment sections, leverage lifetimes for better code quality


neysofu:
At the time of writing, the state of `unsafe` and lifetime usage within the project needs improvements and could benefit from better safety guarantees. We must explore ways to increase our confidence in the safety of all parts of the project, in light of both:

1. Mainnet release;
2. Any future audit work;
3. Removing mental overhead during refactoring and working on new features.

These are especially relevant to crates `svm-sdk-storage`, `svm-types`, and `svm-storage`.

Given these compelling reasons and while this issue is technically not a blocker, I believe it deserves the same amount of attention and thought as if it were one.

Resources, crates, tools and relevant links:

- Fuzzing (#234).
- Running [MIRAI](https://github.com/facebookexperimental/MIRAI) and automatically discover memory safety bugs.
- Previous [discussions](https://www.reddit.com/r/rust/comments/8s7gei/unsafe_rust_in_actixweb_other_libraries/) regarding the management of unsafe code in `actix`. Extremely relevant and lots of tips IMHO.
- [Rust Safety Dance](https://github.com/rust-secure-code/safety-dance) on GitHub can help us get feedback from Rust community experts.

neysofu:
@YaronWittenstein and I agreed that all safety comments should follow the `# Safety` convention.

neysofu:
We've made lots of improvements on this aspect. Code quality remains an issue in some parts of the codebase, but closing this for now.

PR #261, opened 2021-07-07
Removing obsolete FFI code from `svm-storage`

YaronWittenstein:
Since the `Global State` will be re-written in Rust - the FFI interface that SVM originally targeted for `go-spacemesh Global State` has become obsolete.

Related:
* [SMIP: Global-State interface between SVM and go-spacemesh](https://github.com/spacemeshos/SMIPS/issues/20)
* #260


neysofu (approved):
Great cleanup, we are removing lots of nasty code this way. Approved without need for any changes on my part.

PR #265, opened 2021-07-09
Fixed gas

neysofu:
This PR draft tackles the integration of two features within the SVM runtime:

- Fixed gas compiler. This allows users to submit WASM modules with a formal verification of expected pricing.
- SVM-dialect WASM validation. By SVM dialect, we mean a WASM subset that disallows non-terminating programs and opcodes that would compromise the otherwise fully determinstic nature of WASM sandboxing.


Cargo.toml
```diff
@@ -28,32 +28,31 @@ tracing = "0.1"
...
+svm-sdk = { path = "crates/sdk", default-features = false }
+svm-sdk-tests = { path = "crates/sdk-tests" }
+svm-codec = { path = "crates/codec" }
+svm-layout = { path = "crates/layout" }
+svm-storage = { path = "crates/storage", default-features = false }
+svm-runtime = { path = "crates/runtime", default-features = false }
+svm-ffi = { path = "crates/ffi" }
+svm-runtime-ffi = { path = "crates/runtime-ffi", default-features = false }
+svm-abi-layout = { path = "crates/abi/layout" }
+svm-abi-tests = { path = "crates/abi/tests" }
```
neysofu, line 55: Automatic formatter

crates/gas/src/program/mod.rs
```diff
@@ -3,18 +3,21 @@ use parity_wasm::elements::Instruction;
...
+pub use self::exports::Exports;
 
 /// Parsed Wasm Program.
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct Program {
     imports: Imports,
-
     functions: IndexMap<FuncIndex, Vec<Instruction>>,
+    exports: Exports,
```
YaronWittenstein, line 20: sorry for my OCD, can you place it right after `imports`?

crates/gas/src/read.rs
```diff
@@ -68,6 +72,28 @@ fn read_imports<'m>(module: &Module) -> Result<Imports, ProgramError> {
     }
 }
 
+fn read_exports(module: &Module) -> Result<Exports, ProgramError> {
```
YaronWittenstein, line 75: great piece!

crates/runtime-ffi/src/api.rs
```diff
@@ -25,45 +25,30 @@ static DEPLOY_TEMPLATE_RECEIPT_TYPE: Type = Type::Str("deploy-template receipt")
...
-}
-
-macro_rules! vec_to_svm_byte_array {
-    ($ty:expr, $raw_byte_array:expr, $vec:expr) => {{
-        let (ptr, len, cap) = $vec.into_raw_parts();
-
-        to_svm_byte_array!($ty, $raw_byte_array, ptr, len, cap);
-    }};
+#[inline(always)]
+fn maybe_gas(gas_metering: bool, gas_limit: u64) -> Gas {
```
YaronWittenstein, line 29: since we don't use `gas_metering` for now (we initially thought we would),
maybe a better suitable name would be: `gas_enabled` ?

The `Gas::new()` was meant to ease writing tests that don't care (or want to mess) with anything related to the `Gas`

crates/runtime-ffi/src/api.rs
```diff
@@ -25,45 +25,30 @@ static DEPLOY_TEMPLATE_RECEIPT_TYPE: Type = Type::Str("deploy-template receipt")
...
-
-        tracking::increment_live($ty);
-
-        bytes.bytes = $ptr;
-        bytes.length = $len as u32;
-        bytes.capacity = $cap as u32;
-        bytes.type_id = tracking::interned_type($ty);
-    }};
+#[inline(always)]
+unsafe fn data_to_svm_byte_array(
```
YaronWittenstein, line 38: nice!

crates/runtime/src/context.rs
```diff
@@ -23,8 +22,9 @@ pub struct Context {
     app_addr: AppAddr,
 }
 
-// SVM is single-threaded.
-// `Send`, `Sync` and `Clone` are required by `wasmer::WasmerEnv`
+// On SAFETY:
```
YaronWittenstein, line 25: can we use `#SAFETY` instead?

crates/runtime/src/env/default/address_compute.rs
```diff
@@ -1,7 +1,7 @@
 use crate::env::{self, traits};
 
 use env::{ExtSpawnApp, Template};
-use traits::{AppAddressCompute, TemplateAddressCompute};
+use traits::AddressLocator;
```
YaronWittenstein, line 4: I'm not sure the `locator` is the right term.
It's really a computation process here

crates/runtime/src/env/default/memory.rs
```diff
@@ -24,4 +24,6 @@ impl EnvTypes for DefaultMemEnvTypes {
     type AppAddressCompute = default::DefaultAppAddressCompute;
 
     type TemplateHasher = default::DefaultTemplateHasher;
+
+    type Pricer = svm_gas::resolvers::StandardPriceResolver;
```
YaronWittenstein, line 28: For now that's OK. But It seems more likely that will have more `PriceResolver` in the future 
(see the above comment about versioning).

I think it'd be better to inject it as a dependency to the `Runtime`

crates/runtime/src/env/memory/template_store.rs
```diff
@@ -52,7 +50,7 @@ where
 
         hash.and_then(|h| {
             self.bytes
-                .get(&h)
+                .get(h as &[u8])
```
YaronWittenstein, line 53: I've never encountered that. It uses `Deref` here?

crates/runtime/src/env/mod.rs
```diff
@@ -35,16 +37,12 @@ pub use rocksdb::{RocksAppStore, RocksTemplateStore};
...
-pub use traits::{
-    AppAddressCompute, AppStore, TemplateAddressCompute, TemplateHasher, TemplateStore,
-};
+pub use traits::{AddressLocator, AppStore, TemplateHasher, TemplateStore};
 
-/// Runtime types
-pub mod hash;
-use hash::TemplateHash;
+/// Represents an `Template` Hash.
+pub type TemplateHash = [u8; 32];
```
YaronWittenstein, line 45: In most cases I prefer to use a `newtype` since I can extend it with functionality.
However, in this case, it seems reasonable to use alias instead.

crates/runtime/src/env/mod.rs
```diff
@@ -54,13 +52,16 @@ pub trait EnvTypes {
...
 
     /// Compute `App` address type.
-    type AppAddressCompute: AppAddressCompute;
+    type AppAddressCompute: AddressLocator<ExtSpawnApp, Address = AppAddr>;
 
     /// `Template` content Hasher type.
     type TemplateHasher: TemplateHasher;
+
+    /// A pricing engine for templates.
+    type Pricer: PriceResolver;
```
YaronWittenstein, line 64: See my remark above about inject a resolver.

It's unlikely that we'll want to change the way we compute `TemplateHash` or an `Address` of a `Template` or `Account`. But it's almost certain we'll have versioning in regard to `PriceReslover`.

crates/runtime/src/env/traits/address_locator.rs
```diff
@@ -0,0 +1,8 @@
+/// Computes the address of an item.
+///
+/// The algorithm must be deterministic.
+pub trait AddressLocator<T> {
```
YaronWittenstein, line 4: As written above - I don't think that the notion of `Locator` best describes the code.

crates/runtime/src/runtime/outcome.rs
```diff
@@ -1,5 +1,10 @@
 use svm_types::{Gas, ReceiptLog};
 
+/// A wrapper time for `T` that ships with gas information and collected logs.
```
YaronWittenstein, line 3: `Outcome` is the output of the Transaction. 
It can denote a Succeeding Transaction or a Failing one.
In both cases, the `Outcome` will report the amount of Gas used for executing the transaction and collected logs
while running it.

crates/runtime/src/validation/mod.rs
```diff
@@ -1,53 +1,71 @@
...
+/// I32` type signature.
+pub fn validate_svm_alloc(wasm_module: &[u8]) -> bool {
+    let store = crate::new_store();
+    let module = if let Ok(m) = wasmer::Module::new(&store, wasm_module) {
+        m
+    } else {
+        return false;
+    };
+    for export in module.exports() {
+        let expected_type_signature =
```
YaronWittenstein, line 25: seems that it can be extract externally to the `loop`.
I think that `expected_sig` for this local variable would reduce cognitive load. (the shorthand `sig` is used a lot, in this project. Similar to `ty`)

It's worth considering making the `svm_alloc` expected signature a `static`

crates/runtime/src/validation/mod.rs
```diff
@@ -1,53 +1,71 @@
...
+            ExternType::Function(FunctionType::new([wasmer::Type::I32], [wasmer::Type::I32]));
+        if export.name() == "svm_alloc" && export.ty() == &expected_type_signature {
```
YaronWittenstein, line 27: Maybe we can use the new `Exports` under `Program` (inside `svm_gas`) ?

Since it's not really a Gas-Specific logic. Maybe it will be a good idea to extract to a common crate.
(not sure about that)

crates/runtime/src/validation/mod.rs
```diff
@@ -1,53 +1,71 @@
...
 
-    fn feed_event<'a, 'b: 'a>(
-        &mut self,
-        event: Event<'a, 'b>,
-        _module_info: &ModuleInfo,
-        sink: &mut EventSink<'a, 'b>,
-        _source_loc: u32,
-    ) -> Result<(), Self::Error> {
+    let parser = Parser::default();
+    let parser_events = parser.parse_all(wasm_module);
```
YaronWittenstein, line 40: `events` ?

crates/runtime/src/validation/mod.rs
```diff
@@ -1,53 +1,71 @@
...
+    for event_res in parser_events {
```
YaronWittenstein, line 41: maybe we can use `find` here?

crates/runtime/src/vmcalls/alloc.rs
```diff
@@ -4,6 +4,8 @@ use std::convert::TryFrom;
 
 use crate::Context;
 
+/// Statistically allocate `size` bytes within a [`Context`] `ctx`. It returns
+/// the offsed of newly allocated memory.
```
YaronWittenstein, line 8: typo: `offsed`

crates/runtime/src/wasm_store.rs
```diff
@@ -0,0 +1,18 @@
...
+    // we don't run the risk of importing stuff that specific features don't
+    // use (which would cause compiler warnings).
+    if cfg!(feature = "default-cranelift") {
+        use wasmer::{Cranelift, Universal};
+
+        let compiler_config = Cranelift::default();
+        let engine = Universal::new(compiler_config).engine();
+        Store::new(&engine)
+    } else {
+        panic!("No Wasmer backend available.");
```
YaronWittenstein, line 16: good stuff

crates/types/src/receipt/deploy_template.rs
```diff
@@ -55,7 +58,12 @@ impl TemplateReceipt {
         }
     }
 
-    /// Returns the deployed template address. Panics if deploy has failed.
+    /// Returns the deployed template address.
```
YaronWittenstein, line 61: Returns the deployed [`Template`] [`Address`]

crates/runtime/tests/runtime_tests.rs
```diff
@@ -250,6 +249,40 @@ fn runtime_exec_app_with_ctor_fails() {
     ));
 }
 
+#[test]
+fn runtime_spawn_app_without_gas() {
```
YaronWittenstein, line 253: Missing `memory_` prefix to the test name?

Thank you for making this test

YaronWittenstein (changes requested):

crates/codec/src/error.rs
```diff
@@ -1,9 +1,11 @@
+use thiserror::Error;
+
 use std::fmt;
 
 use crate::Field;
 
 #[allow(missing_docs)]
-#[derive(PartialEq, Clone)]
+#[derive(PartialEq, Clone, Error)]
```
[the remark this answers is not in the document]
neysofu: I think so! The nice think about `thiserror` is that it doesn't expose any non-standard `Error` API, it simply derives stuff.

crates/runtime/src/context.rs
neysofu: Isn't `# Safety` for public APIs that expose `unsafe`? This is an internal comment. Just trying to understand what to do in the future.

YaronWittenstein: Actually I don't know 🤷

crates/runtime/src/env/default/memory.rs
neysofu: This will be done while working on the issue that we'll create now, correct? Not in this PR?

crates/runtime/src/env/memory/template_store.rs
neysofu: No, it doesn't use any trait. Converting arrays to slices with `as` is a primitive feature. What I didn't realize, however, is that we don't need this at all. `.get(h)` works fine now that `TemplateHash` is a `type` and not a `struct`.

crates/runtime/src/context.rs
neysofu: Let's use the same everywhere and not complicate our lives unnecessarily

crates/runtime/src/env/mod.rs
neysofu: I should have asked about this and decided together. Lesson learnt :)

crates/runtime/src/context.rs
YaronWittenstein: so `#SAFETY` everywhere?

neysofu: Yes imho. Fixing and resolving.

crates/runtime/src/env/mod.rs
YaronWittenstein: No worries. when I use `newtype` I usually add `#[repr(transparent)]`
But the code base is also using some aliases

It's not a black or white I think

crates/runtime/src/context.rs
neysofu: Any reason why you don't want the space? Grepping maybe?

YaronWittenstein: or `#Safety` everywhere?

neysofu: Let's just do `# Safety` (with a space) which is the one most commonly used rn in the codebase and also what https://rust-lang.github.io/api-guidelines/documentation.html suggests.

YaronWittenstein: > Any reason why you don't want the space? Grepping maybe?

I've never seen this convention. I've only encountered `#SAFETY` or `#Safety`.
So I'm not sure what's the right choice. I'll let you decide.

neysofu: Settled on `# Safety`

crates/types/src/receipt/deploy_template.rs
neysofu: `Address` is not a valid type so the link wouldn't resolve. I'm still rewording this for more clarity :+1:

crates/gas/src/pricing/resolver.rs
```diff
@@ -2,7 +2,7 @@ use crate::Op;
 
 /// Pricing an `Op` or an import function should be part of the consensus protocol.
 /// (It's crucial since otherwise, different network peers will give the same function a different price).
-pub trait PriceResolver {
+pub trait PriceResolver: Clone {
```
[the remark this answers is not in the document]
neysofu: I didn't want to, but I need this in order to wrap it in a `ProgramPricing` struct when calling `ProgramPricing::new()`. Any ideas?

crates/types/src/receipt/deploy_template.rs
YaronWittenstein: My bad

crates/gas/src/pricing/resolver.rs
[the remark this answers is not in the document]
YaronWittenstein: It seems we'll have to substitute the `R` of `ProgramPricing` in favor of `Box<dyn PriceResolver>`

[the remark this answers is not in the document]
neysofu: I don't see how that will solve the `Clone` issue, am I missing something?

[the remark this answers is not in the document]
YaronWittenstein: Does it help?

crates/runtime/src/env/mod.rs
neysofu: Partially addressed by the `V0` rename. Resolving in favor of the conversion above.

crates/gas/src/pricing/resolver.rs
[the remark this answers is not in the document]
YaronWittenstein: Also, I guess we could do as follows:

```rust
impl<R: PriceResolver + 'static> PriceResolver for &R {
}
```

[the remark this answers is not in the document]
neysofu: What do you mean? How should we instantiate a `ProgramPricing` given a `Box<dyn PriceResolver>` without the `Clone` bound?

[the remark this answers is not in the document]
YaronWittenstein: Actually, I think that the `PriceResolver` trait should eliminate the usage of `&self`.
The opcode price should behave like a constant...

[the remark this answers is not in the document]
neysofu: Ok, nevermind. See https://github.com/spacemeshos/svm/pull/265/commits/7b852245ddb23458a6934c7ff355a2d31e72ebbe. I moved the type bound on a `where` for consistency with the codebase and removed `'static` because I don't think it's needed. Is this fine?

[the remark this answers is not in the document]
neysofu: Possibly, but it would be nice to let the trait support a `PriceResolver` that can say, import opcode pricing from a file and stuff like that. I think we should leave it as it is

crates/runtime/src/runtime/outcome.rs
neysofu: Ok, will add this to the doc.

crates/runtime/src/validation/mod.rs
neysofu: Totally missed that it's invariant. Done.

crates/runtime/src/validation/mod.rs
neysofu: Shall I do it now? What do you think?

YaronWittenstein: Probably not, better to open a future issue...

PR #266, opened 2021-07-09
Nuke the `svm-nibble` crate

neysofu:
`svm-nibble` was part of previous ABI design attempts and it's not used anymore within the SVM.


YaronWittenstein:
Generally, it's OK to remove that crate. The reason I've kept that is that I thought it would be useful once the Global-State
component will move to Rust.

Maybe we can add a tag or something such that it will be easier to grab that piece of code in the future (in case we'd want to)

YaronWittenstein (approved):

neysofu:
I agree, let's create a tag so that we can get back to this once we might need it. As of right now, I would prefer to remove as much zombie code as we can.

PR #269, opened 2021-07-11
FFI Imports Removal

YaronWittenstein:
One of the side-effects of moving the Global-State to sit under the Rust side (as opposed to the current one written in Golang at `go-spacemesh`) is that the host functions accessing it will be written in pure Rust. (native host functions with no FFI usage).

This PR removes the FFI imports interface that existed for passing FFI imports from `go-spacemesh` to `go-svm`.
`go-svm` had to propagate than to the SVM Runtime at last. Consequently, the new API for the `runtime-ffi` will be much more simplified.

In order to make the CI pass without implementing the new native host functions functionality, 2 `runtime-ffi` tests have been annotated with`#[ignore]`. This is a stopgap which will be removed upon the next PR.

So the next PR will introduce the new code for serving SVM with native host functions. 
This mechanism will bu used as a way for Running Wasm code under SVM to access the Global-State (or other Node Services potentially in the future).

Note:
* The PR merged the code under #263
**commit**: [`062de28f9705e6c409569764bd8b07539d83c44f`](https://github.com/spacemeshos/svm/pull/263/commits/062de28f9705e6c409569764bd8b07539d83c44f)


crates/runtime-ffi/tests/api_tests.rs
```diff
@@ -1,21 +1,17 @@
...
 
 use std::convert::TryFrom;
 use std::ffi::c_void;
 use std::{collections::HashMap, vec};
 
-use svm_runtime_ffi as api;
-
 use svm_codec::receipt;
-use svm_ffi::{svm_byte_array, svm_env_t, svm_resource_iter_t, svm_resource_t, tracking};
+use svm_ffi::{svm_byte_array, svm_resource_iter_t, svm_resource_t, tracking};
```
neysofu, line 8: Unused imports `svm_resource_iter_t` and `svm_resource_t`.

crates/runtime-ffi/tests/api_tests.rs
```diff
@@ -1,21 +1,17 @@
...
 use svm_layout::FixedLayout;
 use svm_runtime::{testing::WasmFile, vmcalls, Context};
```
neysofu, line 10: Unused imports `Context` and `vmcalls`.

crates/runtime-ffi/tests/api_tests.rs
```diff
@@ -1,21 +1,17 @@
...
-use svm_types::{Address, State, Type, WasmType, WasmValue};
-
+use svm_runtime_ffi as api;
 use svm_sdk::traits::Encoder;
 use svm_sdk::ReturnData;
-
-use maplit::hashmap;
+use svm_types::{Address, State, Type, WasmType, WasmValue};
```
neysofu, line 14: Other unused imports.

crates/runtime/src/lib.rs
```diff
@@ -20,23 +20,19 @@ pub use runtime::create_rocksdb_runtime;
...
 mod storage;
 
-pub use import::ExternImport;
-
 /// Implements `Context`. Used for managing data of running `SVM` apps.
 mod context;
 pub use context::Context;
 
-/// Implements common functionality to be consnumed by tests.
+/// Implements common functionality to be consumed by tests.
```
neysofu, line 33: I think it's best to use inner doc comments (`//!`) for `mod foobar;` rather than outer doc comments (`///`). Basically the same way we have for `vmcalls/mod.rs`.

https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text

crates/runtime/src/lib.rs
```diff
@@ -20,23 +20,19 @@ pub use runtime::create_rocksdb_runtime;
...
 pub mod testing;
 
-/// Implements the `SVM` vmcalls (a.k.a libcalls / hostcalls / syscalls)
```
neysofu, line 39: ...indeed, like this.

neysofu (approved):
Approved! In the future we might want to use `#![deny(dead_code)]` and similar lints in `runtime-ffi/tests` to kill unused imports, unused variables, etc..

Issue #264, opened 2021-07-08
Replacing the term `App` with `Account`


YaronWittenstein:
Originally, the `App` term has been introduced in order to distinguish between two different kinds of Account:

* `Simple` Account
* `Smart-Contract` Account

Recently, the team has voted for shifting towards the `Account Abstraction` (a.k.a `Account Unification`) model.
This decision turns the `App` term to being redundant. 

Moreover, keeping using the term `App` it might cause confusion to people since there is no difference between an `App` and `Account` - all `Account`s are now (for the v0.3) of the same model.

#### This means that we'll use this terminology:

* `Deploy a Template` 
* `Spawn Account` (out of a given `Template`)
* `Execute Account Transaction / Execute Transaction / Call Account`

#### Related:

* https://github.com/spacemeshos/SMIPS/issues/49
* #253

lrettig:
Makes sense, but I still think we may need a term to refer specifically to the _code_ portion of an account, and we may also need to differentiate between classes (e.g., templates) and instances. Saying "execute an account" sounds a bit strange to me. As one point of comparison, Diem/Move has two types of "programs": "modules" and "scripts" (https://developers.diem.com/main/docs/move-modules-and-scripts). I like this terminology.

YaronWittenstein:
An `Account` encapsulates a couple of properties. One of them is the `code`.
Why not just call it `The Account's code`?

Also, an `Account` is an instance of a `Template`

PR #279, opened 2021-07-14
Moving from `App` to `Account` terminology - Part 3

YaronWittenstein:
This is PR number `3` for implementing issue #264


crates/runtime/src/context.rs
```diff
@@ -19,7 +19,7 @@ pub struct Context {
 
     template_addr: TemplateAddr,
 
-    app_addr: AppAddr,
+    app_addr: AccountAddr,
```
neysofu, line 22: Fix this in next PR?

crates/runtime/src/context.rs
```diff
@@ -52,7 +52,7 @@ impl Context {
         gas_limit: Gas,
         storage: AppStorage,
         template_addr: &TemplateAddr,
-        app_addr: &AppAddr,
+        app_addr: &AccountAddr,
```
neysofu, line 55: As above

neysofu (approved):

crates/runtime/src/context.rs
YaronWittenstein: yes

crates/runtime/src/context.rs
YaronWittenstein: yes

PR #285, opened 2021-07-14
Moving from `App` to `Account` terminology - Part 7

YaronWittenstein:
This is PR number `7` for implementing issue #264


crates/codec/src/receipt/call.rs
```diff
@@ -16,19 +16,19 @@
 //!  On success (`is_success = 0`)
 //!  See [error.rs][./error.rs]
 
-use svm_types::CallReceipt;
-
 use std::io::Cursor;
 
+use svm_types::CallReceipt;
+
```
neysofu, line 22: I have been placing `svm` -related imports before `std` this whole time. What convention should we settle on?

neysofu (approved):

crates/codec/src/receipt/call.rs
YaronWittenstein: I've thought the order should be:

1. 3rd-party dependencies
2. `std`
3. Workspace crates
4. current crate

neysofu: Okay, I missed the workspace crates convention. I thought we packed them together with 3rd party's, I will make sense not to :+1:

PR #298, opened 2021-07-16
Use a `PriceResolver` registry with dynamic dispatch

neysofu:
This allows us to add more `PriceResolver`s within a single `Runtime`. It it necessary once we start to experiment with opcode pricing and pricing revisions on-chain.


crates/runtime/src/env/mod.rs
```diff
@@ -73,6 +70,8 @@ where
 {
     accounts: T::AccountStore,
     templates: T::TemplateStore,
+    /// A [`PriceResolverRegistry`] for calculating gas prices.
```
YaronWittenstein, line 73: do we want this comment? (it's a private field)

YaronWittenstein (approved):

crates/runtime/src/env/mod.rs
neysofu: Ups, left it since I initially made it public. Removing it

Issue #243, opened 2021-06-15
Memoize each Template's function pricing


YaronWittenstein:
Right now, we'll have only a few templates deployed upon Genesis and no new ones will be added,

The gas price of any function associated with these templates will be a constant, there is no point of re-computing it.
So memoization of each function price is pretty much a no-brainer.

Once, when SVM will support deploying new Templates - this mechanism will have to be rewritten as a Cache.
(with some eviction strategy such as LRU etc)

PR #299, opened 2021-07-16
Naive template memoization

neysofu:
Solves #243.

### DIFF
--- a/crates/runtime/src/runtime/default.rs
+++ b/crates/runtime/src/runtime/default.rs
@@ -55,32 +55,22 @@ impl<T> DefaultRuntime<T>
 where
     T: EnvTypes,
 {
-    /// Initializes a new [`DefaultRuntime`](DefaultRuntime).
+    /// Initializes a new [`DefaultRuntime`](DefaultRuntime). `template_prices`
+    /// offers an easy way to inject an append-only, naive caching mechanism to
+    /// the [`Template`] pricing logic; using a `None` will result in a new
+    /// empty cache and on-the-fly calculation for all [`Template`]s.
     pub fn new(
         env: Env<T>,
         imports: (String, wasmer::Exports),
         storage_builder: Box<StorageBuilderFn>,
         config: Config,
+        template_prices: Option<Rc<RefCell<HashMap<TemplateAddr, FuncPrice>>>>,
     ) -> Self {
-        let template_prices = Rc::new(RefCell::new(HashMap::default()));
-        Self {
-            env,
-            imports,
-            storage_builder,
-            config,
-            template_prices,
-        }
-    }
-
-    /// Initializes a new [`DefaultRuntime`](DefaultRuntime) with a naive cache
-    /// of [`Template`] [`FuncPrice`]s.
-    pub fn with_template_prices(
-        env: Env<T>,
-        imports: (String, wasmer::Exports),
-        storage_builder: Box<StorageBuilderFn>,
-        config: Config,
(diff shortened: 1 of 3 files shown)


crates/runtime/src/runtime/default.rs
```diff
@@ -42,6 +41,8 @@ where
 
     /// Builds an `AccountStorage` instance.
     storage_builder: Box<StorageBuilderFn>,
+
+    template_prices: HashMap<TemplateAddr, FuncPrice>,
```
YaronWittenstein, line 45: I think we need to make it a `static` variable (outside of `DefaultRuntime`)

We probably don't want to re-use the same `DefaultRuntime` instance throughout the process lifetime.
Doing so implies we trust the Wasm Runtime (currently Wasmer but possible other one) not to have any
internal memory leak, or introduced bugs due to heavy usage.

Creating a new `DefaultRuntime` should be negligible. I think that we are better of with creating a Runtime per usage at this point in time.

YaronWittenstein (changes requested):

crates/runtime/src/runtime/default.rs
```diff
@@ -599,15 +601,24 @@ where
...
-        let gas_mode = template_code_section.gas_mode();
         let template_code = template_code_section.code();
+        let gas_mode = template_code_section.gas_mode();
         let program = Program::new(template_code, false).unwrap();
-        let func_price = {
+
+        // We're using a naive memoization mechanism: we only ever add, never
+        // remove. This means there's no cache invalidation at all. We can
+        // easily afford to do this because the number of templates that exist
+        // at genesis is fixed and won't grow.
```
YaronWittenstein, line 612: Please also comment that once we'll have more `PriceResolver` used the caching keys will consist not only of a `TemplateAddr` but of `PriceResolver Id` (this `id` might be `integer/string` etc).

That way, each `Template` will hold multiple cached `FuncPrice`.
The parameters for resolving a `Price Resolver` will be determined by what we'll decide in the future.
It could be a `LayerId` or the `svm_version` the `Template` has been compiled against (or both `LayerId` and `svm_version`).

In any case, mapping using `PriceResolver Id` will be ignorant to the parameters given to the `PriceResolverRegistry`.

YaronWittenstein (changes requested):

crates/runtime/src/runtime/default.rs
neysofu: I think we should wrap `Runtime`s in a `Context` of some sort. I don't feel comfortable using a global variable...

neysofu: Or maybe pass in a shared `HashMap` to the `Runtime`

YaronWittenstein: That's fine as well to pass to `DefaultRuntime#new` a dependency like:
`cache: &mut HashMap<TemplateAddr, FuncPrice>`

YaronWittenstein: > I think we should wrap `Runtime`s in a `Context` of some sort. I don't feel comfortable using a global variable...

That's understandable. However, it seems that SVM will need some initialization code in order to create the shared dependency (to be injected later for each `DefaultRuntime#new`)

crates/runtime/src/runtime/default.rs
neysofu: See the documentation for method `with_template_prices`

YaronWittenstein (approved):

PR #300, opened 2021-07-16
Revamp the `svm-cli` crate

neysofu:
This PR refactors the `svm-cli` and leverages the crates `clap`, `thiserror` and `anyhow` to provide better documentation and error messages.


crates/cli/src/main.rs
```diff
@@ -1,56 +1,72 @@
...
-    if let Some(wasm_file) = opts.wasm_file {
-        let file = File::open(&wasm_file);
-
-        if file.is_err() {
-            println!("File {:?} doesn't exist", &wasm_file);
-            return;
-        }
+    App::new("smwasm")
+        .version("1.0")
+        .author("The Spacemesh team")
```
YaronWittenstein, line 21: yes, we need to your this term in each crate (I've put my personal name because it was only me working basically on the project until now)

YaronWittenstein (approved):
LGTM!

crates/cli/src/main.rs
neysofu: Np

PR #305, opened 2021-07-19
Some cleaning

YaronWittenstein:
* Removing obsolete code.

* Fixing `cargo doc` warnings.


crates/runtime/src/env/mod.rs
```diff
@@ -50,7 +50,7 @@ pub trait EnvTypes {
     /// [`Template`] type.
     type TemplateStore: TemplateStore;
 
-    /// [`AccountStore`] type.
+    /// `AccountStore` type.
```
neysofu, line 53: Why are we removing the intra-doc links? Am I missing something?

crates/runtime/src/env/traits/mod.rs
```diff
@@ -1,11 +1,11 @@
 //! Runtime traits
 
-mod address_locator;
+mod compute_address;
```
neysofu, line 3: :+1:

neysofu (changes requested):
Looks good! I don't understand why you removed a lot of intra-doc links, that's all. Can you please expand on the reason?

crates/runtime/src/env/mod.rs
YaronWittenstein: The `cargo doc` emitted warnings, probably I'm missing something

neysofu: Seems to be about `mod` visibility... I'm looking for a solution

neysofu: Ok, so the warning is actually a symptom, not a problem itself. We're exposing private `traits` within `EnvTypes`, which is public. `cargo-doc` doesn't like that because of the `rustdoc::private_intra_doc_links` lint. I should we should the public-private `EnvTypes` issue rather than removing intra-doc links.

neysofu (approved):
Much better with `pub` traits! Looks good to me, merge away

PR #306, opened 2021-07-19
Runtime `Context` rename

YaronWittenstein:
The `Context` under `svm-runtime` served as the `Host Function Environment` for each host function.

The name originally was chosen to mirror the naming of `Wasmer` in its older version. (It had a `WasmerCtx` type serving as a general-purpose `Function Environment` to be used by each host function). 

Since then, `Wasmer` has been rewritten and today it ships with a `#[derive(wasmer::WasmerEnv)]` to be annotated by
structs functioning as the `Function Environment`.

This PR would renames `Context` into `FuncEnv`. The reason we don't rename to just `Env` is because there is already an 
`Env` type under `svm-runtime`. Also, the `FuncEnv` conveys better its purpose, to serve as the `Function Environment` to an host function.

Furthermore, the upcoming integration of SVM into the Node will use the term `Context` to denote the `Transaction Context`.
So using the word `Context` in the new AA terminology will refer to the `Transaction Context`.
It will contain fields such as the `current layer`, the referenced `State Root Hash` and similar.


crates/runtime/src/func_env.rs
```diff
@@ -19,10 +19,10 @@ pub struct Context {
 ///
 /// SVM is single-threaded.
 /// `Send`, `Sync` and `Clone` are required by `wasmer::WasmerEnv`.
-unsafe impl Send for Context {}
-unsafe impl Sync for Context {}
+unsafe impl Send for FuncEnv {}
+unsafe impl Sync for FuncEnv {}
```
neysofu, line 23: These two are a hazard, by the way, but we'll address this in a future PR as it predates this change. SVM is indeed single-threaded, but `wasmer` might pass this between threads and bad stuff will happen...

crates/runtime/src/env/mod.rs
```diff
@@ -50,7 +50,7 @@ pub trait EnvTypes {
     /// [`Template`] type.
     type TemplateStore: TemplateStore;
 
-    /// [`AccountStore`] type.
+    /// `AccountStore` type.
```
neysofu, line 53: Reverts the changes from #305, so we should fix this.

neysofu (changes requested):
Looks good, there's no actual changes as it's just search & replace. I would prefer if we fixed the intra-doc links before merging, otherwise we're good.

neysofu (approved):

PR #310, opened 2021-07-20
codec - ASCII Charts slight modifications.

YaronWittenstein:
Slight modifications to ASCII Charts under the `svm-codec` crate.


crates/codec/src/template.rs
```diff
@@ -1,10 +1,12 @@
-//! Encoding for templates
+//! Encoding for binary a [`Template`]s
```
neysofu, line 1: Typo?

neysofu (approved):
Cool ASCII diagrams!

crates/codec/src/template.rs
YaronWittenstein: yes, it'll be fixed in the next PR

Issue #309, opened 2021-07-19
AA: Add `Context` to the API


YaronWittenstein:
This explanation above is a copy-paste from this SMIP:
 [`AA Transactions & SVM Integration SMIP`](https://github.com/spacemeshos/SMIPS/issues/53).

"
In addition to the explicit data sent over-the-wire, there will be implicit fields inferred from it.
One such example, is the `Transaction Id`. For the context of this document, we won't regard these implied fields as part of the `Transaction`.

Adding to a binary `Transaction`, there is the `Execution Context` (or alternatively the `Transaction Context` or simply `Context`). The `Context` structure will contain additional data (alongside the `Transaction`) to be used by SVM when executing a `Transaction`. Properties such as `Layer` or `Block Number` and the `State Root Hash` will be part of it.
It'll be probably useful to attach implied properties such as `Transaction Id` as well to sit under the `Context`.

When SVM will be given a `Transaction` to execute it'll receive both the `Transaction` data and its attached `Context`.
"

Issue #253, opened 2021-06-16
C-API changes


YaronWittenstein:
# Motivation

Since the Global-State will be moved to Rust, the APIs (C and Rust) of SVM will have to adapt accordingly.

Issue #244, opened 2021-06-15
AA:  Add  `Envelope` to the API


YaronWittenstein:
The SVM code as it's today, has no idea of what's a `layer`, `sender`, `nonce` etc.

SVM needs to accept these fields that are part of any transaction (the transaction envelope).
On top of that - SVM will offer new host functions to return part of these fields.

Originally, we target the external imports (living in `go-svm`) to own these fields. 
Since then, it's been decided that SVM will manage the `Global-State` (will be rewritten in Rust) and that each running transaction will be essentially a SVM transaction.

This issue is a natural side-effect of the explained above. The running code will expect these newly added imports to sit under the namespace `sm` and not `svm`. (`sm` is a shorthand for `spacemesh`)

PR #311, opened 2021-07-20
Transaction `Envelope` and `Context`

YaronWittenstein:
This PR attempts to be faithful to the [`AA Transactions & SVM Integration SMIP`](https://github.com/spacemeshos/SMIPS/issues/53).
It's a perquisite for starting the integration with `go-svm` and later with the `go-spacemesh`.

**Closes** #253 #244 #309

## Changes

* Added `Envelope` and `Context` to `svm-types`.
* Started using the term `Message` when referring to the binary SVM transaction under `svm-runtime` and `svm-runtime-ffi`.
* Implemented encoding for `Envelope` and `Context` under `svm-codec`.
* Modified the `Runtime` trait under `svm-runtime` to use the `Envelope` and `Context`.
* Simplified creation of `In-memory Runtime` (doesn't require an injected `Key-Value` anymore).
* Adapted the tests of `svm-runtime` according to the above changes.
* Modified the `runtime-ffi` API to adapt to the SMIP as well.
* Renamed previously missed usages of `App`  into `Account`.

## Future Issues:

* #313 
* #314


crates/runtime/src/runtime/default.rs
```diff
@@ -563,78 +562,106 @@ impl<T> Runtime for DefaultRuntime<T>
 where
     T: EnvTypes,
 {
-    fn validate_deploy(&self, bytes: &[u8]) -> std::result::Result<(), ValidateError> {
-        let template = self.env.parse_deploy(bytes, None)?;
+    fn validate_deploy(&self, message: &[u8]) -> std::result::Result<(), ValidateError> {
+        let template = self.env.parse_deploy(message, None)?;
```
neysofu, line 566: Good job on changing this variable name, we should start doing this more often. Rather than `bytes`, `slice`, let's use stuff such as `envelope`, `payload`, `wasm_module`, etc.. More descriptive stuff!

crates/runtime/src/runtime/default.rs
```diff
@@ -681,55 +707,30 @@ where
...
-        //             let v: &wasmer::Val = returns.first().unwrap();
-
-        //             v.i32().unwrap() == 0
-        //         });
-
-        //         out.map_err(|fail| fail.take_error())
-        //     } else {
-        //         unreachable!("Should have failed earlier when doing `validate_tx`");
-        //     }
+        todo!("https://github.com/spacemeshos/svm/issues/248")
```
neysofu, line 714: Finally this will be out of our way eheh. Good job

crates/runtime/src/runtime/mod.rs
```diff
@@ -35,39 +32,34 @@ use crate::error::ValidateError;
...
-    ///
-    /// On Success:
-    /// * Persists changes to the called `Account`'s storage.
-    /// * Receipt returns the `Account`'s new `State`.
-    /// * Receipt returns the amount of [`Gas`] used.
-    ///
-    /// On failure:
-    /// * Receipt returns the occurred error
-    /// * Receipt informs the amount of gas used (transaction gas limit)
-    fn call(&self, tx: &Transaction, state: &State, gas_limit: Gas) -> CallReceipt;
```
neysofu, line 72: You think this is outdated and better suits the documentation of `CallReceipt` itself?

crates/types/src/transaction/context.rs
```diff
@@ -0,0 +1,60 @@
+use crate::{Layer, State, TransactionId};
+
+/// Holds the `Transaction`'s Context.
+///
+/// Once created it should NOT be modified (immutable).
+/// It contains properties related to the context in which a Transaction will execute.
+/// Additionally, it will encapsulate properties inferred from the binary Transaction sent over-the-wire.
```
neysofu, line 7: Good summary. I suggest once change though: Rather than saying that you shouldn't modify `Context`, let's say that you can't. Period. Something like "An immutable container for the `Transactions`'s Context. ... This type does not allow any modifications as that would ... (insert bad stuff here, why we can't allow users to do it)".

As of now, it seems that you *can* modify it even though you shouldn't, while in reality there's no methods that let you do that.

Does this make sense?

crates/types/src/transaction/envelope.rs
```diff
@@ -0,0 +1,82 @@
+use crate::{Address, Gas};
+
+/// Holds `Transaction` **agnostic** content.
+///
+/// Once created it should NOT be modified (immutable).
```
neysofu, line 5: Same thing... possibly? WDYT?

crates/types/src/transaction/layer.rs
```diff
@@ -2,3 +2,9 @@
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[repr(transparent)]
 pub struct Layer(pub u64);
+
+impl Default for Layer {
+    fn default() -> Self {
+        Self(0)
+    }
+}
```
neysofu, line 10: Can you please explain to me why 0 is the `Default` `Layer`? I think I'm missing something

neysofu: BTW we might want to turn this into a `type` alias in the future.

crates/runtime/tests/runtime_tests.rs
```diff
@@ -247,98 +212,57 @@ fn memory_runtime_call_func_not_found() {
...
-    // 2) `Spawn Account`
-    let name = "My Account";
-    let ctor = "initialize";
-    let calldata = vec![];
-    let creator = Address::of("creator").into();
-    let bytes = testing::build_spawn(&template_addr, name, ctor, &calldata);
-    let receipt = runtime.spawn(&bytes, &creator, Gas::with(0));
-
-    assert!(matches!(receipt.error.unwrap(), RuntimeError::OOG));
-}
```
neysofu, line 281: What happened to this one?

neysofu (changes requested):
Small stuff to change, I've given you details above. Otherwise looks good, awesome so see that we're moving closer to integration.

crates/codec/src/context.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use crate::{ReadExt, WriteExt};
+
+/// Encodes a binary [`Context`] of a transaction.
+pub fn encode(context: &Context, w: &mut Vec<u8>) {
+    w.write_tx_id(context.tx_id());
+    w.write_u64_be(context.layer().0);
+    w.write_state(context.state());
+}
+
```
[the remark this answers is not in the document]
YaronWittenstein: Honestly, I don't think we should do it.

I prefer right to keep the types agnostic to any encoding concerns,
and have the crate more lean. 

The `svm-types` is ubiquitous whilst the encoding not necessarily. 
The `codec` is pretty much for encoding of Transactions (and a bit of `runtime-ffi` encoding).

crates/ffi/src/byte_array.rs
```diff
@@ -64,6 +64,16 @@ impl svm_byte_array {
...
+
+    /// Returns a byte slice over the contents.
+    pub fn as_slice(&self) -> &[u8] {
+        unsafe { std::slice::from_raw_parts(self.bytes, self.length as usize) }
+    }
+
+    /// Copies `self` into a new `Vec`.
+    pub fn to_vec(&self) -> Vec<u8> {
+        self.as_slice().to_vec()
+    }
```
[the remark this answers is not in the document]
YaronWittenstein: so I've renamed the old `new` to `pub fn with_capaicity` and added a `pub unsafe fn new(..)`

crates/ffi/src/byte_array.rs
```diff
@@ -117,16 +127,13 @@ impl TryFrom<&svm_byte_array> for String {
...
         // data is cloned here, so the new `String` won't be merely an alias,
         // and `bytes` will still require a separate de-allocation.
         //
         // Making it an alias is unsafe because the data may not have
         // been dynamically allocated, or not by Rust's global allocator.
-        let vec = slice.to_vec();
 
-        String::from_utf8(vec)
+        String::from_utf8(bytes.to_vec())
     }
```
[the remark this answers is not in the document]
YaronWittenstein: Having the `unsafe new` isn't not enough?

crates/types/src/transaction/context.rs
YaronWittenstein: Yes, also `Context` is written as an immutable type.

crates/types/src/transaction/layer.rs
YaronWittenstein: > Can you please explain to me why 0 is the `Default` `Layer`? I think I'm missing something

Couldn't think of any other value to use, when we don't really care about the `Layer`.
We could maybe use `Option<..>` but it seems over-kill in this case IMHO.

YaronWittenstein: > BTW we might want to turn this into a `type` alias in the future.

Probably we could. However,  sometimes we might end up wanting to implement a trait or two, and we won't be able to do
that when using an `alias` type.

crates/runtime/src/runtime/mod.rs
YaronWittenstein: Yes. It's a bit verbose, not complete with some incorrect info.
I would keep the new short description for now.

crates/codec/src/context.rs
```diff
@@ -0,0 +1,41 @@
...
+/// Returns the decoded [`Context`],
+/// On failure, returns [`std::io::Result`].
+pub fn decode(cursor: &mut Cursor<&[u8]>) -> std::io::Result<Context> {
+    let tx_id = cursor.read_tx_id()?;
+    let layer = cursor.read_u64_be()?;
+    let state = cursor.read_state()?;
+
+    let context = Context::new(tx_id, Layer(layer), state);
+    Ok(context)
+}
```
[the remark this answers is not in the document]
YaronWittenstein: As I've said, I'd really prefer to keep the `types` from encoding in isolation for now.

A good example from the current code base is the `svm-storage` crate.
It's using the `svm-types` as a depedendency.
However, It doesn't rely on the `svm-codec` (since it's Transaction & Receipts focused).

Similarly, the new Global-State will probably use `svm-types` and it'll not have any reason I can currently imagine 
to use the `svm-codec`.

The `svm-codec` is more for communication with the external (out of SVM) world.
It might be Transactions, Receipts or some `runtime-ffi` (In-memory buffers encoding).

crates/runtime/tests/runtime_tests.rs
YaronWittenstein: https://github.com/spacemeshos/svm/issues/313

crates/runtime/src/runtime/mod.rs
neysofu: Ok!

crates/ffi/src/byte_array.rs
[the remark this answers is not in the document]
neysofu: Cool, now this UB is impossible to have so we can resolve this.

crates/codec/src/context.rs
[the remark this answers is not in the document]
neysofu: Ok then, np :+1: just an idea.

crates/types/src/transaction/layer.rs
neysofu: I agree on the `type` vs `struct` issue after some thinking.

crates/ffi/src/byte_array.rs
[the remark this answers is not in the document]
neysofu: It is :+1:

crates/types/src/transaction/layer.rs
neysofu: I don't think `Layer` should have a `Default`. Semantically, it doesn't make much sense. Considering that it's only used for tests, how about we add `#[cfg(test)]` to `Context::with_state` and fill it with dummy data (0 is fine, just like you said)?

```rust

    /// Creates a new [`Context`] with the given `state` parameter.
    ///
    /// Sets dummy values for all remaining fields.
    ///
    /// # Notes
    ///
    /// This function/method is only available during testing.
    #[cfg(test]
    // (We can also move this to `runtime_tests.rs`)
    pub fn with_state(state: State) -> Self {
        Self {
            tx_id: TransactionId::zeros(),
            layer: Layer::default(),
            state: state,
        }
    }
```

neysofu (approved):

PR #312, opened 2021-07-20
Refactoring of `svm_codec::api::json` with Serde

neysofu:
At the moment, `svm_codec::api::json` uses `serde_json::Value` for most internal JSON manipulation. This is relatively error-prone compared to the `#[derive(Serialize, Deserialize)]` alternative, which allows Serde to automatically figure out the encoding details of your data structure.

This PR makes the following changes:

- Change most JSON-related public functions to take raw strings as inputs rather than `serde_json::Value`. This hides Serde types from users and allows us to leverage `serde_path_to_error`.
- Simplify error messages while making sure to give full paths to the JSON location where the error appeared.
- Create some internal utility data structures within `serde_types.rs` that can be used to serialize and deserialize lots of SVM transaction data.


YaronWittenstein (approved):
Very nice!

PR #316, opened 2021-07-25
Refactoring for SDK Auto-generated Metadata

YaronWittenstein:
Compiling a `#[template]` program using the SVM SDK emits (when target is `Wasm`) SVM flavored Wasm code.

In others words, the emitted Wasm conforms to the restrictions of SVM:

* Each import is covered by a corresponding host function.
* Wasm is Fixed-Gas compliant.
* Floats are prohibited.

However, the Wasm is not the whole story.
In order to produce a valid `Deploy` transaction we need to encode these sections at a minimum:

* `Code` Section - It contains the Wasm code, but additional fields as well.
* `Ctors` Section - It contains a list of `Exports` considered a `constructor` (can be used only for `Spawn`-ing Accounts).
* `Data` Section - It specifies the Template's storage variables layout.

This PR puts us one step closer to having a `Deploy Template` generation from existing SDK `#[template]`.
Initially, I thought it will be easiest if the `svm-sdk-macros` crate will add the `svm-codec` as a dependency. 
This seemingly trivial idea turned out to be tricky.

The `svm-codec` depends on a few SDK crates using the `dynamic-alloc` feature flag.
The reason that the `codec` don't use the `static-alloc` feature flag is since it's not required for it.
The `static-alloc` is part of the `Fixed-Gas` mechanism and it assumes that the running code is a SVM programs.
Even when the `codec` is executed as a Wasm program (to be consumed by `smapp` or similar), it's **not** a valid SVM program (and it's never meant to be one).

On the other side, each SDK crate used by `svm-codec`, is also used by `svm-sdk-macros` but this time with its `static-alloc` feature flag. Consequently, trying to add `svm-codec` as a dependency to `svm-sdk-macros` will cause **feature-flags conflicts.**
The `static-alloc` and `dynamic-alloc` features **can't** co-exist.

The fact that the `svm-codec` is used solely for the Procedural-Macros implementation (i.e at compile-time) doesn't solve the issue. There is no way to tell the compiler that the scope of `svm-codec` is restricted to Procedural-Macros only.

It seems that the simplest and most elegant solution will be to extract the `Deploy` transaction into other crate.
The `svm-cli` seems the best place to do that.

The current `svm-sdk-macros` already emits JSON files containing all the required information for proceeding with this task.
This PR refactors the code and introduces different terms. Instead of using the term `Schema` from now, we'll use the term `meta` (see also `TemplateMeta` in the code). Also the `api` and `layout` feature flags have been replaced by a single `meta` feature flag.

Additionally, the `meta` feature flag is used only under the `lib.rs` file. Tests under `svm-sdk-macros-tests` will turn-off that flag.
This PR contains a skeleton of code at file `sections.rs` under `svm-cli` (most code is commented out).

A next PR should introduce a new sub-command for constructing binary `Deploy` templates.
This sub-command will receive as inputs:
 
* A file path for the `#[template]` compiled Wasm.
* A file path for the `meta` JSON emitted by using the SDK (with the `meta` feature flag turned-on).

This sub-command will craft a binary `Deply Transactions` using the `svm-codec` crate.
For now, it'll suffice to construct **only the mandatory** Sections (`Code`, `Data` and `Ctors`).


neysofu (approved):
Looks good!

PR #317, opened 2021-07-26
Catch panics within critical FFI calls

### LINKED ISSUE #233: use `std::panic::catch_unwind` in each `runtime-ffi` API
Panicking from Rust to C is undefined behavior.
By wrapping each such call with `std::panic::catch_unwind` we can fail safely.

neysofu:
Closes #233.


YaronWittenstein (approved):
LGTM!

Issue #301, opened 2021-07-19
Tracking issue for transaction crafting using `svm-cli`


neysofu:
Ideally, users should be able to use `svm-cli` for crafting transactions of different types and with custom data. This requires adding a subcommand for each transaction type (`spawn`, `deploy`, `call`).

- [x] #303
- [x] #302
- [x] #304
- [ ] Generic cleanup after `svm_codec::api::json` refactorings.

PR #318, opened 2021-07-26
Craft Deploy transactions from the CLI

neysofu:
This is the continuation of PR #316, which was preparatory for the API hereby implemented. We provide a high-level API that, given the outputs of the SVM SDK, can easily craft Deploy transaction binaries.

Closes https://github.com/spacemeshos/svm/issues/301.


crates/cli/src/subcmd_craft_deploy/meta.rs
```diff
@@ -0,0 +1,77 @@
...
+
+#[derive(Debug, Serialize, Deserialize)]
+struct TemplateMetaSchema {
+    id: u64,
+    name: String,
+    #[serde(rename = "type")]
+    ty: String,
+    offset: usize,
+    byte_count: usize,
+    length: Option<usize>,
```
YaronWittenstein, line 52: It's not ideal to turn `length` into `Option.
I'm not asking to change that now, but ideally it's be an enum

crates/cli/src/subcmd_craft_deploy/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+                .takes_value(true),
+        )
+}
+
+pub fn subcmd_craft_deploy(args: &ArgMatches) -> anyhow::Result<()> {
+    let smwasm_code = {
+        let path = args.value_of("smwasm").unwrap();
+        std::fs::read(path)?
+    };
+    let smwasm_meta: TemplateMeta = {
```
YaronWittenstein, line 50: I usually prefer shorter names like `code` or `wasm` instead of `smwasm_code`.
Similarly, `meta` instead of `smwasm_meta`.

However, it's up to you to decide whether to rename or not.

crates/cli/src/subcmd_craft_deploy/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+        let path = args.value_of("meta").unwrap();
+        let string = std::fs::read_to_string(path)?;
+        serde_json::from_str(string.as_str())?
+    };
+
+    let code_section = {
```
YaronWittenstein, line 56: I personally find it a bit harder to read than:

```rust
let flags = CodeSection::exec_flags();
let code_section = CodeSection::new(
  svm_types::CodeKind::Wasm,
  smwasm_code,
  flags,
  svm_types::GasMode::Fixed,
  0,
)
};
```

Usually, I try to minimize nesting.

crates/cli/src/subcmd_craft_deploy/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+        CodeSection::new(
+            svm_types::CodeKind::Wasm,
+            smwasm_code,
+            flags,
+            svm_types::GasMode::Fixed,
+            0,
+        )
+    };
+
+    let encoded = {
```
YaronWittenstein, line 67: same, as above. This is easier to read IMHO (no nesting and a space between `sections` building to `encoder`)

```rust
let mut sections = Sections::with_capacity(3);
sections.insert(Section::Code(code_section));
sections.insert(Section::Ctors(smwasm_meta.ctors_section()));
sections.insert(Section::Data(smwasm_meta.data_section()));

let mut encoder = SectionsEncoder::with_capacity(3);
encoder.encode(&sections);
let encoded = encoder.finish()
```

p.s: the `codec` generally uses `bytes` instead of `encoded`.
It's up to you to decide whether to keep it as such or not.

crates/cli/src/subcmd_craft_deploy/meta.rs
```diff
@@ -0,0 +1,77 @@
...
+                builder.push(schema_var.byte_count as u32);
+            }
+        }
+
+        DataSection::with_layout(svm_layout::Layout::Fixed(builder.build()))
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct TemplateMetaSchema {
```
YaronWittenstein, line 45: `TemplateMetaVar` ?

crates/cli/src/subcmd_craft_deploy/meta.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct TemplateMetaApiSignature {
+    params: Vec<TemplateMetaApiSignatureParam>,
+    returns: Json,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct TemplateMetaApiSignatureParam {
```
YaronWittenstein, line 72: `MetaSigParam / MetaSignatureParam` ?

crates/cli/src/subcmd_craft_deploy/meta.rs
```diff
@@ -0,0 +1,77 @@
...
+    name: String,
+    wasm_name: String,
+    is_ctor: bool,
+    is_fundable: bool,
+    doc: String,
+    signature: TemplateMetaApiSignature,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct TemplateMetaApiSignature {
```
YaronWittenstein, line 66: `MetaSig / MetaSignature` 

I think that `Template` and `Api` are meaningless in this context.

crates/cli/src/subcmd_craft_deploy/meta.rs
```diff
@@ -0,0 +1,77 @@
...
+    schema: Vec<TemplateMetaSchema>,
+    api: Vec<TemplateMetaApi>,
+}
+
+impl TemplateMeta {
+    pub fn ctors_section(&self) -> CtorsSection {
+        let ctors = self
+            .api
+            .iter()
+            .filter(|api_entry| api_entry.is_ctor)
```
YaronWittenstein, line 24: can you rename `api_entry` to `export`? 
Probably using the term `Exports` instead of `API` under `svm-codec` would be more precise.

YaronWittenstein (dismissed):

crates/cli/src/subcmd_craft_deploy/meta.rs
neysofu: I agree, this is not super elegant, we should change this in the future.

PR #319, opened 2021-07-26
Removing `AddressOf<..>`

### LINKED ISSUE #314: Deprecate the usage of `AddressOf<..>`
Use `Address` or `newtypes` wrapping `Address`
This should reduce noise from the code and make it more clean

YaronWittenstein:
Removing `svm_types::AddressOf` which used has been used as an `Address` container.
The new code is more simplified and reduces noise. 

We're only left now with two types:

* `Address` - Represents an `Account Address`
* `TemplateAddr` - Represents a `Template Address`

## Note

Right now `TemplateAddr` is configured as `20 bytes` - the same as `Address`.
We haven't decided yet about the byte-length of a `Template Address`.
It's likely it will stay `20 bytes`, but we will keep these two distinct types regardless.

Closes #314


neysofu (approved):
I think having `AccountAddr` would improve readability but it's just my preference. LGTM, approved.

PR #321, opened 2021-07-27
Singlepass feature flag

YaronWittenstein:
Adding the `default-singlepass` feature-flag that will use the Wasmer's SinglePass Compiler.

This should result in faster compilation times for macOS/Linux and thus a better development experience of SVM.
The CI will still compile using the `default-cranelift` since Cranelift compiles on Windows as well.

If the Singlepass Compiler will support Windows at some point in the future, we'll probably use it for production as well
(and stop using Cranelift altogether).

This PR also removes the `default-llvm` feature flag. LLVM is a very heavy dependency and it's unlikely that SVM will ever really use it.


crates/runtime/src/wasm_store.rs
```diff
@@ -1,18 +1,21 @@
...
+}
+
+/// New fresh `Store`.
+#[cfg(feature = "default-singlepass")]
+#[must_use]
+pub fn new_store() -> Store {
+    use wasmer::{Singlepass, Universal};
+
+    let engine = Universal::new(Singlepass::default()).engine();
+    Store::new(&engine)
```
neysofu, line 20: Why the change? The function is defined with both `default-singlepass` and `default-cranelift`; what changes is the implementation body. I think the old code better reflected that.

neysofu (changes requested):

crates/runtime/src/wasm_store.rs
YaronWittenstein: The old code would result in a compilation error:
```
"could not find `Cranelift` in `wasmer`"
```

PR #323, opened 2021-07-28
svm-runtime: Injecting the `Envelope` and `Context` into `FuncEnv`

YaronWittenstein:
This PR implements injection of both the `Envelope` and `Context` into the `FuncEnv`.
This is a preparation work for adding future host functions that will expose more information.

As an example, an executed transaction can ask for it's `Id` (the `transaction id` sits under the `Context`).


neysofu (approved):
LGTM!

PR #324, opened 2021-07-29
`svm-runtime-ffi` tests restoration

YaronWittenstein:
This #269 PR made the `svm_runtime_success` made `svm_runtime_failure`  obsolete.

This PR re-implements both tests under the new design. 

### Notes:

* Since there is no option to inject imports using the FFI layer anymore, the Rust examples using the SVM SDK have been updated as well.

* Both `svm_envelope_alloc` and `svm_context_alloc` have been modified not to accept any input.
The binary size of `Envelope` and `Context` is fixed.

* When encoding an `Error Message` as part of a `Receipt`, the message is truncated to be of length 255 bytes at most.
(the classic example, is having an error message containing a deep call-stack).


neysofu (changes requested):
Some changes I'd like to see:

- Remove `unsafe` from `svm_envelope_alloc`, `svm_message_alloc`, `svm_context_alloc`. They're not needed (not introduced with this PR but since we're at it we might as well do it).
- Remove `svm_byte_array::bytes`. It's equivalent to `x.as_slice().as_ptr()`, which is a `safe` alternative.
- Replace `svm_byte_array::bytes_mut` with a safe `fn as_slice_mut(&mut self) -> &mut [u8] { ... }`. Note the added safety and how it requires `&mut`.
- Safer `byte_array_copy`:

```
fn byte_array_copy(byte_array: &mut svm_byte_array, bytes: &[u8]) {
    byte_array.as_slice_mut().copy_from_slice(bytes);
}
```
This reflects in tests under `api_tests.rs` e.g.:

```
let mut byte_array = api::svm_message_alloc(msg.len() as u32);
byte_array_copy(&mut byte_array, &msg);
```

PR #331, opened 2021-08-02
Adding `returndata byte-size` field to Receipts

YaronWittenstein:
Implementation of a binary Receipt decoding should not necessarily be aware to the ABI encoding used for the `returns` of the transaction (a.k.a the `returndata`). 

In other words, upon a successful `spawn/call` transaction, the Receipt decoder code should be able to decode its fields and treat the `returndata` as a black-box (i.e a blob of data). 

This PR prepends each `returndata blob` with 2-bytes (Big-Endian as always) signaling the `returndata` total size in bytes.
I've noticed this necessity while writing the [Receipts Encoding](https://github.com/spacemeshos/SMIPS/issues/58) SMIP.


neysofu (approved):
LGTM!

PR #332, opened 2021-08-03
AA `verify`

YaronWittenstein:
This PR implements the `verify` stage as part of the Runtime interface.
More in-depth information about why need that functionality appears in the  [Account Unification] SMIP.

## Notes

* We use the term `Verify Data` to denote the ABI encoded data given as input for the `verify` function.
* We use the term `Call Data` to denote the ABI encoded data given when running the Transaction function.
* Since both `VerifyData` and `CallData` share the same mechanism, parts of the code simply use the term `Input Data`.
In order not to bloat this PR, there are still code parts that use the term `Call Data` but gradually they will be renamed to use `Input Data`.
* Right now the `verify` returns a `CallReceipt`. It means that `verify` can return any blob of data (a.k.a the `returndata`).
Whether the `verify` will return only a boolean is still in research.
*  For now, the `verify` is expected to work with no access to the Account's storage. In other words, only pure logic
can take place within `verify`. See  issue: #327 

[Account Unification]: https://github.com/spacemeshos/SMIPS/issues/49


crates/program/src/lib.rs
```diff
@@ -2,10 +2,10 @@
...
 //! The main entity in this crate is [`Program`].
 
-#![deny(missing_docs)]
-#![deny(unused)]
-#![deny(dead_code)]
-#![deny(unreachable_code)]
+#![allow(missing_docs)]
+#![allow(unused)]
+#![allow(dead_code)]
+#![allow(unreachable_code)]
```
neysofu, line 8: Do we need all of these set to `allow`?

neysofu (approved):
Looks good to me :+1:

crates/program/src/lib.rs
YaronWittenstein: Pushed a fix

Issue #252, opened 2021-06-16
Global-State implementation


YaronWittenstein:
We've decided to move the current Global-State component from `go-spacemesh` into the Rust

The core of SVM will consume its API and it should facilitate significantly the integration of SVM into the Full-Node.
The `go-svm` library will have to undergo the required changes and would become much more simple.

For now, we want to do the bare minimum required in order to achieve that. So using off-the-shelf solution such as [`parity-trie`] must be considered before writing everything from scratch.

In the future, we might rewrite this component to suit better our future needs (for example, using the new code to implement dynamic data-structures for SVM apps)

- [x] #328 (not necessary for 0.3 thanks to LinkedHashX)
- [x] #329 (not necessary for 0.3 thanks to LinkedHashX)
- [x] #330 
- [x] #358 

[`parity-trie`]: https://github.com/paritytech/trie

neysofu:
We can close this :rocket:

PR #333, opened 2021-08-04
Global State: versioned KV store with root signature calculation

neysofu:
This is a PR draft for the versioned key-value data store foundation upon which we're building the Global State. Root signature is calculated via XORing leaf hashes (https://eprint.iacr.org/2021/773.pdf, pg. 5-6) rather than MPT. This is known as LinkedHashX and we might be able to replace Merklelization walks while we don't need partial verification.

See https://github.com/spacemeshos/svm/issues/252.


crates/hash/Cargo.toml
```diff
@@ -11,4 +11,5 @@ description = "Spacemesh Virtual Machine"
 publish = false
 
 [dependencies]
-tiny-keccak = "1.4.2"
\ No newline at end of file
+blake3 = "1"
+tiny-keccak = { version = "2", features = ["keccak"] }
```
YaronWittenstein, line 15: I think we can rely solely on `Blake3` throughout the codebase.

crates/hash/src/lib.rs
```diff
@@ -1,24 +1,81 @@
...
+
+    fn finalize(self) -> Self::Hash {
+        *self.hasher.finalize().as_bytes()
+    }
 }
 
 /// Implements the `KeyHasher` trait using the `keccak256` hashing algorithm (output: 32 bytes)
-pub struct DefaultHasher;
+pub struct DefaultHasher {
+    hasher: tiny_keccak::Keccak,
```
YaronWittenstein, line 38: As said above, probably better to replace with `Blake3`

crates/state/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "svm-state"
+version = "0.1.0"
+edition = "2018"
+publish = false
```
YaronWittenstein, line 5: Please make the `[package]` section similar to the other crates

crates/state/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1"
+blake3 = "1"
+funty = "=1.1.0"
+futures = "0.3"
+redis = { version = "0.21", features = ["tokio-comp"] }
```
YaronWittenstein, line 14: Why redis?

crates/state/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+edition = "2018"
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1"
+blake3 = "1"
+funty = "=1.1.0"
+futures = "0.3"
```
YaronWittenstein, line 13: Do we really need Futures here?

crates/state/src/error.rs
```diff
@@ -0,0 +1,19 @@
+use std::fmt::Debug;
+
+pub type Result<T> = std::result::Result<T, GlobalStateError>;
+
+#[derive(Debug, thiserror::Error)]
+pub enum GlobalStateError {
+    #[error("Invalid item.")]
+    InvalidItem,
```
YaronWittenstein, line 8: Can you please put one line between each item (I find it a bit distraction due to the proc-macros used)

crates/state/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+///
+/// * https://www.programmersought.com/article/33363860077/
+/// * https://www.youtube.com/watch?v=oEpY4NkkeYQ
+///
+/// Lists of namespaces within [`GlobalState`]:
+///
+/// * 't': templates
+///   * ''
+/// * 'a': accounts
+///   * 'b': balance
```
YaronWittenstein, line 41: Why do we need a namespace for `balance`, `nonce`?

crates/state/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+///   * 'n': nonce
+///   * 'a': address
+pub struct GlobalState {
+    sqlite: SqlitePool,
+    next_commit_id: i64,
+    dirty_changes: HashMap<[u8; 32], Vec<u8>>,
```
YaronWittenstein, line 47: I think it will be better to wrap the `[u8; 32]` inside a `newtype`

crates/state/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+impl GlobalState {
+    /// Creates a new [`GlobalState`] persisted by the given SQLite instance.
+    pub async fn new(sqlite_uri: &str) -> Result<Self> {
+        let sqlite = sqlx::SqlitePool::connect(sqlite_uri).await?;
+        let mut gs = Self {
+            sqlite,
+            dirty_changes: HashMap::new(),
+            next_commit_id: 1,
+        };
+        gs.initialize().await?;
```
YaronWittenstein, line 59: I don't see any reason to use `async` in the Global-State API

crates/state/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+    /// persisted changes without dirty changes.
+    pub async fn current(&self) -> Result<Commit> {
+        self.commit_signature(self.next_commit_id).await
+    }
+
+    /// Returns the current root hash in the form of a [`Commit`].
+    ///
+    /// It returns an error in case there's any dirty changes: you must
+    /// [`rollback`](GlobalState::rollback) first.
+    pub async fn rewind(&mut self, _commit: &Commit) -> Result<()> {
```
YaronWittenstein, line 293: I think that the rewind should do something like:

```rust
pub fn rewind(&mut self, commit_id: u64) {
   assert!(commit_id <= self.next_commit_id);

  self.next_commit_id = commit_id + 1;
}
```

We need the Global-State to be able to rewrite an existing `commit_id`.
If using the auto-increment feature collides with that requirement, maybe we can make it work manually.
(implementing the increment of the `commit_id` ourselves).

Since the client of the Global-State is known to work in a linear order I think it shouldn't be hard.

crates/state/src/lib.rs
YaronWittenstein: Using `async` internally against SQLite is OK of course

crates/state/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+    /// Sets the `value` associated with a Blake3 `hash`. This change will be "dirty" until
+    pub async fn upsert_by_hash<V>(&mut self, hash: [u8; 32], value: V)
+    where
+        V: Into<Vec<u8>>,
+    {
+        let value = value.into();
+        self.dirty_changes.entry(hash).or_insert(value);
+    }
+
+    pub async fn merkelize(&self) -> Option<Commit> {
```
YaronWittenstein, line 188: Do we still need it?

crates/state/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+    /// [`GlobalState::upsert_by_hash`] for more information.
+    pub async fn upsert<V>(&mut self, key: &[u8], value: V)
+    where
+        V: Into<Vec<u8>>,
+    {
+        let hash = Blake3Hasher::hash(key);
+        self.upsert_by_hash(hash, value).await;
+    }
+
+    /// Sets the `value` associated with a Blake3 `hash`. This change will be "dirty" until
```
YaronWittenstein, line 179: The comment seems to require more explanation

crates/state/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+                "#,
+                self.next_commit_id,
+                key_hash,
+                value
+            )
+            .execute(&self.sqlite)
+            .await?;
+        }
+
+        self.update_commit_signature(&signature).await
```
YaronWittenstein, line 245: If we are only inserting new `(key, value, next_commit_id) why do we want to call the `update_commit_signature`
?

YaronWittenstein (changes requested):
I've given some comments - Let's continue iterating on this PR.

crates/hash/src/lib.rs
YaronWittenstein: Please add to `lib.rs`:

```rust
#![deny(missing_docs)]
#![deny(unused)]
#![warn(dead_code)]
#![deny(unreachable_code)]
#![feature(vec_into_raw_parts)]
```

crates/state/src/lib.rs
YaronWittenstein: Please add to `lib.rs`:

```rust
#![deny(missing_docs)]
#![deny(unused)]
#![warn(dead_code)]
#![deny(unreachable_code)]
```

crates/state/Cargo.toml
YaronWittenstein: Ok, it seems not to be used (the same for `funty` and `futures`)

crates/state/src/lib.rs
YaronWittenstein: The higher level API will probably have a similar:

```rust
pub fn rewind(&mut self, layer: u64) {
}
```

So a `layer` is essentially a `commit_id`

crates/state/src/resources/schema.sql
```diff
@@ -0,0 +1,31 @@
...
+	"id" INTEGER NOT NULL UNIQUE,
+	"signature" BLOB NOT NULL CHECK(length("signature") == 32),
+	PRIMARY KEY("id")
+);
+
+DROP TABLE IF EXISTS "values";
+CREATE TABLE IF NOT EXISTS "values" (
+	"id" INTEGER NOT NULL UNIQUE,
+	"commit_id"	INTEGER,
+	"key_hash" BLOB NOT NULL,
```
YaronWittenstein, line 14: It's a shame we can't tell SQLite that `key_hash` is of a fixed-size.

crates/state/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+        }
+    }
+
+    /// Erases all dirty changes from memory. Persisted data is left untouched.
+    pub fn rollback(&mut self) {
+        self.dirty_changes.clear();
+    }
+}
+
+#[cfg(test)]
```
YaronWittenstein, line 307: It's important that we simulate:

* Rewinding back to some `commit` and replaying the same transactions as previously but in a different order.
* Rewinding back to some `commit` and playing new transactions.
* Playing two transactions successfully, then fail the 3rd one and after that succeed the 4th transaction (when the 4th and 3rd transactions share part of the keys).

YaronWittenstein (changes requested):

crates/state/src/lib.rs
neysofu: Exactly, it's mandatory.

crates/state/Cargo.toml
neysofu: Yeah, my bad

crates/state/Cargo.toml
neysofu: `funty` is needed to avoid breakage with `nom` and `bitvec` (see https://github.com/myrrlyn/funty/issues/3). I've moved `futures` to `[dev-dependencies]`.

`redis` indeed is not needed, a remnant of some old prototyping.

crates/state/Cargo.toml
neysofu: Addressed above.

crates/state/src/lib.rs
neysofu: I'll use `Fingerprint` as Noam suggested, I think it fits the bill.

crates/state/src/lib.rs
neysofu: Nope, we don't. You're right!

crates/state/src/resources/schema.sql
neysofu: I'll add a `check`, that's the best we can do.

crates/state/src/lib.rs
neysofu: On this point, we might need to make some decisions regarding forcing linear ordering at the API level. We do support querying of historical data, but rewinding and then playing new transactions requires soft-deleting the newer commits or something along that lines. This is one limitation we have, which shouldn't be a problem for reorgs (as you said, clients only use linear behavior) but it is for the kind of test you're describing. I'll think about this and get back to you.

crates/state/src/lib.rs
neysofu: Any specific reason why we don't want to write to SQLite on `checkpoint`? Saving state on `checkout` is consistent with what `git` and other systems generally mean by that word. What does a `Journal` (or `dirty_tx_changes` / `dirty_commit_changes`) achieve?

crates/state/src/lib.rs
neysofu: Just examples and fillers for now.

crates/state/src/lib.rs
neysofu: Auto-incremental shouldn't cause any problems, I think we can leave it as it is! If you were to delete some commit ID, then you'd have a "hole". But that doesn't cause issues with range queries and similar.

crates/state/src/lib.rs
neysofu: Solved, we're not writing anything to db when using `checkpoint`.

YaronWittenstein (approved):
Great job!

PR #334, opened 2021-08-10
Protected Mode for `verify`

YaronWittenstein:
This PR restricts storage access while running the `svm_verify` and `svm_alloc` functions.

* When running `svm_alloc` there is no reason to have access to the storage of an Account.
* When running `svm_verify` we won't it to executed logic to be stateless.
More information under the [Account Unification] SMIP.

[Account Unification]: https://github.com/spacemeshos/SMIPS/issues/49


neysofu (approved):
LGTM!

PR #351, opened 2021-08-23
Safer C APIs

neysofu:
See #260.

- Use `#![deny(unsafe_op_in_unsafe_fn)]`, which makes auditing `unsafe` code much easier.
- Use method-based casting on raw pointers rather than `as`. Behavior is better documented and they protect against `NULL` deference.
- Try to push `unsafe` code as high as possible within function bodies to highlight what `unsafe` operations we do.
- Cast pointers to reference as early as possible as they are the main cause of `unsafe`.


YaronWittenstein (approved):
Very nice!

PR #352, opened 2021-08-23
Disallow broken intra-doc Cargo links

neysofu:
As per title. This is also checked by CI, which ensures that we can always build documentation for `master`.


YaronWittenstein (approved):
Great!

PR #354, opened 2021-08-24
Only compile svm-runtime-ffi in CI

neysofu:
Workaround for https://github.com/rust-lang/rust/issues/50007.


YaronWittenstein (approved):
Looks good!

PR #357, opened 2021-08-30
Global State - Higher level APIS (Part 1)

neysofu:
First PR in a series of PRs that will phase out `svm-storage` and replace it with `svm-state`.


crates/state/src/gs.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    pub async fn in_memory() -> Self {
+        Self {
+            storage: Storage::in_memory().await.unwrap(),
+        }
+    }
+
+    pub async fn account_balance(&self, account_addr: &Address) -> Result<Option<u64>> {
+        let key = format!("accounts:{}:balance", account_addr.to_string());
```
YaronWittenstein, line 25: might be better to use shorter encoding here.
also I think that the Account data should be mapped by a single key

crates/state/src/gs.rs
```diff
@@ -0,0 +1,153 @@
...
+            let s = std::str::from_utf8(&bytes[..]).expect("Invalid UTF-8 value.");
+            let balance = str::parse(s).expect("Invalid numeric value.");
+            Ok(Some(balance))
+        } else {
+            Ok(None)
+        }
+    }
+
+    pub async fn account_nonce(&self, account_addr: &Address) -> Result<Option<u64>> {
+        let key = format!("accounts:{}:nonce", account_addr.to_string());
```
YaronWittenstein, line 38: same comment as `account_balance`

crates/state/src/gs.rs
```diff
@@ -0,0 +1,153 @@
...
+            let s = std::str::from_utf8(&bytes[..]).expect("Invalid UTF-8 value.");
+            let balance = str::parse(s).expect("Invalid numeric value.");
+            Ok(Some(balance))
+        } else {
+            Ok(None)
+        }
+    }
+
+    pub async fn template_address(&self, account_addr: &Address) -> Result<Option<TemplateAddr>> {
+        let key = format!("accounts:{}:template_address", account_addr.to_string());
```
YaronWittenstein, line 51: same comment as `account_balance` and `account_nonce`

crates/state/src/gs.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        Ok(opt_value.map(|x| TemplateAddr::new(&x[..])))
+    }
+
+    pub async fn template_mandatory_sections(
+        &self,
+        account_address: &Address,
+    ) -> Result<Option<TemplateAddr>> {
+        let key = format!(
+            "templates:{}:mandatory_sections",
```
YaronWittenstein, line 62: using shorter keys is not critical since there will be significantly less Templates in comparison to Accounts.
But for staying under the same convention - might be better to shorter the encoding `t:{}:ma`

crates/state/src/gs.rs
```diff
@@ -0,0 +1,153 @@
...
+        let opt_value = self.storage.get(key.as_bytes(), None).await?;
+
+        Ok(opt_value.map(|x| TemplateAddr::new(&x[..])))
+    }
+
+    pub async fn template_optional_sections(
+        &self,
+        account_address: &Address,
+    ) -> Result<Option<TemplateAddr>> {
+        let key = format!(
```
YaronWittenstein, line 74: same comment about encoding keys as in `template_core_sections`

crates/state/src/gs.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    pub async fn set_template_optional_sections<V>(
+        &mut self,
+        template_addr: &TemplateAddr,
+        value: V,
+    ) -> Result<()>
+    where
+        V: Into<Vec<u8>>,
+    {
+        let key = format!("templates:{}:optional_sections", template_addr.to_string());
```
YaronWittenstein, line 91: I think it will be better to extract keys construction into a function (to avoid DRY)

crates/state/src/gs.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    pub async fn set_template_mandatory_sections<V>(
+        &mut self,
+        template_addr: &TemplateAddr,
+        value: V,
+    ) -> Result<()>
+    where
+        V: Into<Vec<u8>>,
+    {
+        let key = format!("templates:{}:mandatory_sections", template_addr.to_string());
```
YaronWittenstein, line 104: same comment. can be extracted into a utility function for DRY avoidance

crates/state/src/gs.rs
```diff
@@ -0,0 +1,153 @@
...
+        self.storage.upsert(key.as_bytes(), value.into()).await;
+        Ok(())
+    }
+
+    pub async fn set_account_balance(
+        &mut self,
+        account_addr: &Address,
+        balance: u64,
+    ) -> Result<()> {
+        let key = format!("accounts:{}:balance", account_addr.to_string());
```
YaronWittenstein, line 114: same comment. can be extracted into a utility function for DRY avoidance

YaronWittenstein (changes requested):

crates/state/src/gs.rs
neysofu: Right, I forgot that we're using SQLite now. No key compression at all. I'll absolutely reduce key size :+1:

neysofu:
I don't think we need to shorten the keys. They get hashed into 32 bytes anyway. Hashing a a few bytes more with Blake3 literally only takes a few CPU cycles, so...

crates/state/src/global_state.rs
```diff
@@ -0,0 +1,293 @@
...
+    // -----------------
+
+    pub async fn template_sections(
+        &self,
+        template_addr: &TemplateAddr,
+    ) -> Result<Option<Sections>> {
+        let core_sections_opt: Option<Sections> = self
+            .read_and_decode(&keys::template_core(template_addr))
+            .await?;
+        let noncore_sections_opt: Option<Sections> = self
```
YaronWittenstein, line 172: same comment as the last one

crates/state/Cargo.toml
```diff
@@ -14,8 +14,12 @@ publish = false
 anyhow = "1"
 blake3 = "1"
 futures = "0.3"
-sqlx = { version = "0.5", features = ["runtime-tokio-rustls", "sqlite", "macros"] }
+# https://github.com/launchbadge/sqlx/issues/713
```
YaronWittenstein, line 17: We discussed over Slack that it would probably be better to revisit in the future whether we could avoid using `Actix` and use `Tokio`

YaronWittenstein (changes requested):

crates/state/Cargo.toml
neysofu: https://github.com/spacemeshos/svm/issues/365 indeed!

crates/state/src/global_state.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    // TEMPLATE SECTIONS
+    // -----------------
+
+    pub async fn template_sections(
+        &self,
+        template_addr: &TemplateAddr,
+    ) -> Result<Option<Sections>> {
+        let core_sections_opt: Option<Sections> = self
```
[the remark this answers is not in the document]
neysofu: :+1:

crates/state/src/global_state.rs
neysofu: Fixed!

PR #372, opened 2021-09-11
Flattening `crates/`: better Cargo workspace structure

neysofu:
This is a follow up PR to #371.

It flattens the `crates/` directory which, as far as I can tell, is more in line with Rust practices and it allows us to use wildcards within the top-level `Cargo.toml`. Testing-only crates are now part of `tests/`.


YaronWittenstein (approved):

neysofu:
Made obsolete by https://github.com/spacemeshos/svm/pull/379.

PR #387, opened 2021-09-24
Provide static libraries as artifacts

neysofu:
As per title. This might be useful for `go-svm`.


YaronWittenstein (approved):
Awesome job!

PR #393, opened 2021-10-04
codec - removing the `Deploy Template` from the Wasm and JSON APIs

YaronWittenstein:
As we've agreed, the process for generating a Template will be:

1. Developing a Template using the SVM SDK - the compiler will emit a Wasm file and JSONs.
2. Taking the outputs of 1. and feeding them into the SVM CLI (`craft deploy`).
3. The CLI will emit the so-called `Deploy Message` 

@avive FYI


neysofu (approved):
:+1:

avive:
I already pulled the updated lib for the npm package.

PR #411, opened 2021-11-12
Run `cargo upgrade --workspace`

neysofu:
Let's settle on using `cargo upgrade` to automagically manage dependencies' versions, rather than picking through them manually.


kobby-pentangeli (approved):
Running `cargo upgrade --workspace` does the magic, indeed. It works!

Another command in our arsenal is `cargo update [_options_]`. This command will update dependencies in the `Cargo.lock` file to the latest version. A `Cargo.lock` file with the latest versions of the deps will be created if one doesn't exist.

PR #412, opened 2021-11-12
Safety improvements: mitigation for PR #390

neysofu:
This is a follow-up PR to https://github.com/spacemeshos/svm/pull/411.

https://github.com/spacemeshos/svm/pull/390 fixed a Fixed-Gas regression in `svm-abi-decoder`, meaning that some code had been merged into `master` which caused the compiled `.wasm` artifacts to contain illegal `loop` instructions. This broke SVM Fixed-Gas mode, so we reverted back to using pointers within `Cursor`.

This PR mitigates the two issues:
- It doesn't compile into any `loop` instructions and invalid WASM code.
- It reverts *some* of the safety improvements applied before breaking Fixed-Gas. In particular, we introduce back explicit lifetimes, which help the compiler check for some useful invariants and reduces the UB surface.

PR #416, opened 2021-11-19
Safety improvements: (additional) mitigation for PR #390

neysofu:
This is a follow-up PR to #412. It reverts almost all of the original safety improvements, with additional care not to break Fixed-Gas invariants. The use of `unsafe` is localized and all usage of pointers in `Cursor`-related public interfaces is replaced with slices.


kobby-pentangeli (approved):
Good work, @neysofu. I only suggest we `.gitignore` the following files:

1. `CallDataTemplate-meta.json`, 
2. `Cargo.lock` (especially since this is a library and not a binary), and 
3. `runtime_calldata.wasm`.

... unless there are explicit reasons to keep them in the history.

neysofu:
@kobby-pentangeli I agree with `.gitignore`-ing `runtime_calldata.wasm` and `CallDataTemplate-meta.json`. They are artifacts, and as such they have no place in the source repository. I'll create an issue right now, I feel it's best addressed in a separate PR but it's a good point.

The SVM project is neither a library nor a binary, rather a Rust workspace which contains both libraries and binaries. The official [docs](https://doc.rust-lang.org/cargo/reference/workspaces.html) suggest using a shared `Cargo.lock`, so I'm inclined to keep it.

crates/abi-decoder/src/cursor.rs
```diff
@@ -39,8 +39,8 @@ impl<'a> Cursor<'a> {
         if self.is_eof() {
             Option::None
         } else {
-            let byte = unsafe { *self.offset_ptr() };
-            Option::Some(byte)
+            let byte = unsafe { self.bytes.get_unchecked(self.offset) };
```
YaronWittenstein, line 42: @neysofu IIRC, you said that using the `get_unchecked` makes the emitted Wasm code Fixed-Gas Wasm compliant. Otherwise, I don't have any other comment to add to this PR, and you can go ahead and merge it!

PR #418, opened 2021-11-19
Use Rust 2021 edition (new feature resolver!)

neysofu:
This is a follow-up PR to #417. Now that most Cargo feature conflicts are gone (only some work on `static-alloc` and `dynamic-alloc` is missing, but these two feature flags turned out to be the less problematic among all, so I've prioritized `ffi` and `mock`), we can use the 2021 Rust edition without issues.


kobby-pentangeli (approved):
Since this is a separate PR, I'm reiterating this suggestion here as well: 

Unless there are explicit reasons otherwise, I suggest we `.gitignore` the following files:
1. `CallDataTemplate-meta.json`, 
2. `Cargo.lock` (especially since this is a library and not a binary), and 
3. `runtime_calldata.wasm`.

YaronWittenstein (approved):

Issue #406, opened 2021-10-31
Implement the `svm_transfer` host function


YaronWittenstein:
Add a new host function that will send coins from the current executing account to a destination account.

The host function should live under:
https://github.com/spacemeshos/svm/tree/master/crates/runtime

## Notes

* The code should integrate with the `AccountStorage` (wraps the `Global State`):
https://github.com/spacemeshos/svm/blob/master/crates/state/src/account_storage.rs
* If the destination account doesn't exist, then the host function should **panic**.
This behavior might change in case the `Self Spawn` feature is implemented.

### Testing the Host Function in isolation

Under the `svm-runtime` there're tests that exercise the host functions without creating a full-blown SVM Runtime.

Here is the file containing such tests:
https://github.com/spacemeshos/svm/blob/master/crates/runtime/tests/vmcalls_tests.rs

Now, in order to generate Wasm inputs for the tests, please look at the current ones:
https://github.com/spacemeshos/svm/tree/master/crates/runtime/tests/wasm

### Other Tests

* Testing a `Template` running within the SVM Runtime

This is not a requirement for that issue. But it's something we'll want to have.
One way to achieve that will be by extending this test:

https://github.com/spacemeshos/svm/blob/478d12fddfa87e6230a64010bfd3c0f24ad3164f/crates/runtime/tests/runtime_tests.rs#L311

* Testing the SDK - See issue: #407

neysofu:
Closed by #419.

PR #419, opened 2021-11-22
Implementation of `svm_transfer` host function

kobby-pentangeli:
This PR addresses [issue#406](https://github.com/spacemeshos/svm/issues/406), i.e. introducing a new host function that sends coins from the current executing account to a destination account.

TODO:
- Investigate why loading the source account succeeds, but doing same for the destination account fails.


neysofu:
Reviewing :)

neysofu:
The loading of the destination account fails because the helper function `create_account` in `vmcalls_tests.rs` creates a new `GlobalState::in_memory` at every call. The two accounts effectively exist in two separate databases. I believe we've never had a test that required creating multiple accounts!

The solution is:
1. Add a `gs: &GlobalState` argument to `create_account` and `.clone()` it inside the function body if necessary. Cloning a `GlobalState` is relatively cheap because it's wrapped in an `Arc`.
2. Put `let gs = GlobalState::in_memory();` where necessary and add it as an argument to `create_account`.

crates/runtime/src/vmcalls/transfer.rs
```diff
@@ -0,0 +1,34 @@
...
+///
+/// # Panics
+///
+/// Panics when the destination account does not exist.
+pub fn svm_transfer(env: &FuncEnv, src_addr: i32, dst_addr: i32, amount: i64) {
+    let borrow = env.borrow();
+    let storage = borrow.storage();
+
+    let src_addr = Address::repeat(src_addr as u8);
+    let dst_addr = Address::repeat(dst_addr as u8);
```
YaronWittenstein, line 15: This is not correct.

YaronWittenstein (changes requested):

crates/runtime/src/vmcalls/transfer.rs
moshababo: Can you further elaborate?

YaronWittenstein: `Address::repeat` takes a `u8` and creates a new `Address` having 20 bytes looking the same - 
It's useful for testing. 

The `src_addr` and `dst_addr` should have acted as pointers to a Memory cell (pointing to the first byte of an `Address`)

crates/runtime/src/runtime/runtime.rs
```diff
@@ -707,6 +707,34 @@ impl Runtime {
...
+    /// # Panics
+    ///
+    /// Panics when the destination account does not exist.
+    pub fn transfer(&self, src_addr: &Address, dst_addr: &Address, amount: u64) {
+        let mut src_account = AccountStorage::load(self.gs.clone(), src_addr).unwrap();
+
+        let mut dst_account = if let Some((_bal, _counter, _addr)) = self.get_account(dst_addr) {
+            AccountStorage::load(self.gs.clone(), dst_addr).unwrap()
+        } else {
+            panic!("Destination account does not exist")
```
YaronWittenstein, line 722: @neysofu this soon will have to be modified to create a `Stub Account` as detailed in the [`Self-Spawn SMIP`](https://github.com/spacemeshos/SMIPS/issues/64)

crates/runtime/src/vmcalls/transfer.rs
```diff
@@ -0,0 +1,46 @@
+use crate::FuncEnv;
+use svm_state::AccountStorage;
+use svm_types::{Address, BytesPrimitive};
+
+fn load_addr(env: &FuncEnv, mem_ptr: u32) -> Address {
```
YaronWittenstein, line 5: I'd extract the code that reads bytes out of memory using const generics and have the `load_addr` code receive `[Address::len(); u8]` and only make an `Address.`

IIRC there's some API for that built-in in Wasmer (for grabbing a Wasm memory bytes)

crates/runtime/src/vmcalls/transfer.rs
```diff
@@ -0,0 +1,46 @@
...
+/// Sends coins from the current executing account to a destination account.
+///
+/// # Panics
+///
+/// Panics when the destination account does not exist.
+pub fn svm_transfer(env: &FuncEnv, src_addr_ptr: i32, dst_addr_ptr: i32, amount: i64) {
+    let borrow = env.borrow();
+    let gs = borrow.storage().gs.clone();
+
+    let src_addr = load_addr(env, src_addr_ptr as u32);
```
YaronWittenstein, line 26: The `load_addr` could receive `i32` or `WasmPtr` instead of `u32`

crates/runtime/src/vmcalls/transfer.rs
```diff
@@ -0,0 +1,46 @@
...
+    let src_bal = src_account.balance().unwrap();
+    let dst_bal = dst_account.balance().unwrap();
+
+    let amount = amount as u64;
+
+    if src_bal < amount {
+        panic!("Not enough balance to execute transfer")
+    }
+    src_account
+        .set_balance(src_bal.checked_sub(amount).unwrap())
```
YaronWittenstein, line 41: The `checked_sub` seems redundant due to the `src_bal < amount` check above.
It's true that code changes but I'd just make sure that this logic is tested.

crates/runtime/src/vmcalls/transfer.rs
```diff
@@ -0,0 +1,46 @@
...
+    let borrow = env.borrow();
+    let gs = borrow.storage().gs.clone();
+
+    let src_addr = load_addr(env, src_addr_ptr as u32);
+    let dst_addr = load_addr(env, dst_addr_ptr as u32);
+
+    let mut src_account = AccountStorage::load(gs.clone(), &src_addr).unwrap();
+    let mut dst_account = AccountStorage::load(gs, &dst_addr).unwrap();
+
+    let src_bal = src_account.balance().unwrap();
```
YaronWittenstein, line 32: typo? (same for `dst_bal`)

crates/runtime/tests/vmcalls_tests.rs
```diff
@@ -354,3 +358,74 @@ fn vmcalls_log() {
...
+        include_str!("wasm/svm_transfer.wast").into(),
+    );
+
+    let func: NativeFunc<(u32, u32, i64)> =
+        instance.exports.get_native_function("transfer").unwrap();
+    let src_addr_ptr = 0u32;
+    let dst_addr_ptr = 20u32;
+    let amount = 100i64;
+
+    for (cell, byte) in memory
```
YaronWittenstein, line 409: Please consider extracting a future issue that Tests will have a utility to copy a slice of bytes into an existing Instance. (maybe Wasmer have already something like that as well)

YaronWittenstein (changes requested):

crates/runtime/src/runtime/runtime.rs
neysofu: Yup, will have to open a PR soon! For now we're just panicking.

crates/runtime/src/vmcalls/transfer.rs
neysofu: `WasmPtr` is nicer, didn't think about it. Fixing now.

crates/runtime/src/vmcalls/transfer.rs
neysofu: Just short variable naming leftover from the old code, fixing now.

crates/runtime/src/runtime/runtime.rs
neysofu: https://github.com/spacemeshos/svm/issues/424.

crates/runtime/tests/vmcalls_tests.rs
neysofu: https://github.com/spacemeshos/svm/issues/425.

PR #423, opened 2021-11-29
Invalid `ctor`: `FuncNotAllowed` is now `FuncNotFound`

neysofu:
When spawning with an invalid `ctor`, we previously returned a `FuncNotAllowed` error. The reasoning was, it's possible that the selected function exists but isn't actually allowed to run as a `ctor`. I think `FuncNotFound` makes more sense though.


neysofu:
I'll close this, @YaronWittenstein intends on doing a more comprehensive refactoring also fixing this issue.

Issue #375, opened 2021-09-12
Return `Touched Accounts` inside the `Spawn` and `Call` Receipts


YaronWittenstein:
#### Notes

Before working on this issue, please make sure this one is addressed #406 

### Details

While calling an existing `Account` (i.e., executing a `Call` transaction) and spawning new ones (involves running a constructor), we might transfer coins between different Accounts. 

`go-spacemesh` would like to be notified which accounts were involved in coins transfers as part of executing a transaction.

In other words, each time the `svm_transfer` host function is being called, we'd like to append the `source` and `destination` into the `Touched Accounts` set.

The logged `source` will always be the same Address - the currently executing Account.
So the implementation can optimize that, but keeping things simple and stupid in the solution might be a better approach.

The task of the issue is to make sure that `Spawn` and `Call` receipts will have a new `touched_accounts` field that will contain this list of `Account Addresses.`

### Encoding

Lastly, this field should also be encoded in the returned Receipts. 

See the `svm-codec` code here:
https://github.com/spacemeshos/svm/tree/master/crates/codec/src/receipt

neysofu:
Just clarifying this: "touched accounts" is a set, and as such doesn't make any distinction between `source` and `destination` because it's, duh, a set (like `std::collections::HashSet`). `source` and `destination` are out-of-band terms that we use to understand what items should be included in that set. @YaronWittenstein do we agree on this?

YaronWittenstein:
To my understanding, yes!
It's the list of accounts that have been involved in at least one coins transfer during the transaction execution.

neysofu:
Nice, we're on the same page then.

PR #430, opened 2021-12-02
Touched accounts

neysofu:
Closes https://github.com/spacemeshos/svm/issues/375.

This PR adds a `touched_accounts` field to both `SpawnReceipt` and `CallReceipt`. This is a set of account addresses that includes the following:

- Principal account address.
- Newly spawned account address (only for `spawn`).
- Any other address involved in a coin transfer during smart contract execution.


YaronWittenstein (approved):
LGTM!!

PR #435, opened 2021-12-09
Runtime refactoring

YaronWittenstein:
I want to refactor the `Runtime` of SVM.

Here are the changes done on this one:

#### Adding `RuntimeError::FuncNotCtor`

The `FuncNotAllowed` requires a `target`.
Since we don't really have a `target` when spawning, I've ended up the `FuncNotCtor`

#### Removal of `Imports` dependency

Initially, the `Runtime` expected external imports.
Back then, we thought that `Global-State`-related host functions would be implemented within the `go-spacemesh`.
Since then, the `Global State` has been moved to the Rust land, so we ended up with having setting empty `imports` as a parameter when initializing the `Runtime`

#### Moved `Template Pricing` to another file

I created `TemplatePriceCache` to act as the cache for the `Template Pricing`. 
The `Runtime` expects now that cache as a dependency upon initialization. 
In practice, it'd be ideal to have only a single cache used throughout the process lifetime.

### Added `GasTank`

I thought it would make the code more readable if the repeating pattern of checking whether we've enough gas during will be less explicit and noisy in the code.

I've also added two helpers: `check_gas_for_payload` and `check_gas_for_func`.
Both expect a paramere named `gas_left` and its type of `GasTank`.
Each helper checks first whether there is any `gas in the tank` and returns immediately if the tank is empty.

The `GasTank` could have been an `Option<u64>,` but I think it's nicer as it's now.

#### Other

* Renamed `ProtectedMode` to `AccessMode` (also added `ImmutableOnly` option for the future).
* Removed in the `exec` function the creation of `AccountStorage` for the `target` - it seems unnecessary. If the target doesn't exist, we should fail (to be done in a separate PR)
* Added more `TODO` under `src/runtime/runtime.rs`
* Might be better to review `src/runtime/runtime.rs` in the editor.

### DIFF
--- a/crates/runtime-ffi/src/api.rs
+++ b/crates/runtime-ffi/src/api.rs
@@ -8,7 +8,7 @@ use std::panic::UnwindSafe;
 use std::slice;
 
 use svm_codec::Codec;
-use svm_runtime::{PriceResolverRegistry, Runtime, ValidateError};
+use svm_runtime::{PriceResolverRegistry, Runtime, TemplatePriceCache, ValidateError};
 use svm_state::GlobalState;
 use svm_types::{Address, BytesPrimitive, Context, Envelope, Layer, State};
 
@@ -68,7 +68,8 @@ pub unsafe extern "C" fn svm_runtime_create(
         }
         *initialized = true;
 
-        let imports = ("sm".to_string(), wasmer::Exports::new());
+        // TODO: move both `GlobalState` and `TemplatePriceCache` to sit under `Env`.
+        // `Env` be a singleton living throughout the process' lifetime.
         let global_state = if path.is_null() {
             GlobalState::in_memory()
         } else {
@@ -76,7 +77,8 @@ pub unsafe extern "C" fn svm_runtime_create(
             GlobalState::new(std::str::from_utf8(db_path).expect("Invalid UTF-8 path."))
         };
 
-        let runtime = Runtime::new(imports, global_state, PriceResolverRegistry::default());
+        let registry = PriceResolverRegistry::default();
+        let runtime = Runtime::new(global_state, TemplatePriceCache::new(registry));
 
         *runtime_ptr = RUNTIME_TRACKER.alloc(runtime);
         debug!("`svm_runtime_create` end");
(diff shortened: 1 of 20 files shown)


crates/runtime-ffi/src/api.rs
```diff
@@ -555,8 +552,16 @@ pub unsafe extern "C" fn svm_create_account(
     catch_unwind_or_fail(|| {
         let runtime = RUNTIME_TRACKER.get(runtime_ptr).unwrap();
         let account_addr = Address::new(slice::from_raw_parts(addr, Address::N));
+        let template_addr = TemplateAddr::god_template();
         let counter = ((counter_upper_bits as u128) << 64) | (counter_lower_bits as u128);
-        runtime.create_account(&account_addr, "".to_string(), balance, counter)?;
+
+        runtime.create_account(
```
neysofu, line 558: `Runtime::create_account` implies that the template address is `TemplateAddr::god_template()` (or rather, whatever we'll end up using for Simple Coin Transfer). Passing is as an argument gives the impression that it's customizable, but it's not. What do you think about this?

crates/runtime/src/func_env.rs
```diff
@@ -28,7 +28,7 @@ impl FuncEnv {
         context: &Context,
         template_addr: TemplateAddr,
         target_addr: Address,
-        mode: ProtectedMode,
+        mode: AccessMode,
```
neysofu, line 31: I like the rename :+1:

crates/runtime/src/func_env.rs
```diff
@@ -118,17 +117,22 @@ pub struct Inner {
...
-pub enum ProtectedMode {
+pub enum AccessMode {
     /// Access to [`AccountStorage`] is not allowed.
     AccessDenied,
 
+    /// Only `Read Access` to [AccountStorage]'s `Immutable Storage` is allowed.
+    ImmutableOnly,
+
     /// Full-Access to [`AccountStorage`] is allowed.
     FullAccess,
```
neysofu, line 137: I think we can make it more consistent: `None`, `Read`, `ReadAndWrite`

neysofu (approved):
Just some minor comments. LGTM!

crates/runtime/src/func_env.rs
YaronWittenstein: You can still have a method granting access only for reading `Immutable State` and disallowing reading other Sections.

neysofu: Right, that would be `Read` (akin to `ImmutableOnly`), right?

YaronWittenstein: Yes

neysofu (approved):

